PR #36, opened 2024-06-30
Miscellaneous fixes to project files

crazeejeeves:
Contains the following minor changes/fixes:
1. Added Models/Output symlink to .gitignore. Useful when using multiple applications sharing models and output folders. Existing gitignore only excludes actual Models/Output folder and not symlinks (which are seen as files instead)
2. There were a few extra unclosed span tags in t2i page, now removed.
3. There was one span tag closed with </button>. This is now fixed.

Note, for (2) and (3), browser DOM manager seem to safely 'fix' this at runtime, so a code change was not strictly necessary. However, this fix removes the errors from polluting the IDE analysis results.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ bin/
 obj/
 .vs/
 .vscode/
+.idea/
 *.user
 *.suo
 Properties/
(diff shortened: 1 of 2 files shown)


mcmonkey4eva:
`Added Models/Output symlink to .gitignore. Useful when using multiple applications sharing models and output folders. Existing gitignore only excludes actual Models/Output folder and not symlinks (which are seen as files instead)`
ftr you can just set an alternate path for these folders under Server Config, you don't need a symlink in the case of swarm

Looks like good fixes though, thank you

PR #37, opened 2024-07-01
#Windows - Add the `/b` switch to `exit` calls in `install-windows.bat`

maedtb:
The `/b` switch tells `exit` to exit the current script normally, but not close the command prompt if (and only if) it was a freestanding process.  This fixes errors in the windows install script from closing the command prompt if it did not launch the window.

If the command prompt was launched to run the script (by clicking on the `install-windows.bat` file in `Explorer`, for example), the window will continue to close normally after this change.

### DIFF
--- a/install-windows.bat
+++ b/install-windows.bat
@@ -5,13 +5,13 @@ cd /d "%~dp0"
 if exist SwarmUI (
     echo SwarmUI is already installed in this folder. If this is incorrect, delete the 'SwarmUI' folder and try again.
     pause
-    exit
+    exit /b
 )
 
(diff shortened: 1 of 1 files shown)


mcmonkey4eva:
... huh?

mcmonkey4eva:
That's what the `pause` line right above the `exit` is for

mcmonkey4eva:
Per Discord discussion, this is an obscure bug workaround for legacy windows 10 and before era `cmd.exe` that won't affect 99% of users, but doesn't hurt to have, so will pull

PR #45, opened 2024-07-03
timer used to debounce user input into the filter: this reduces redundant back end calls by ensuring the user has paused typing before activating.

brendanhoar:
Tested using both Edge and Firefox.

Please let me know if you'd like changes, etc.

### DIFF
--- a/src/wwwroot/js/genpage/browsers.js
+++ b/src/wwwroot/js/genpage/browsers.js
@@ -535,10 +535,15 @@ class GenPageBrowserClass {
                 depthInput.parentElement.style.display = 'none';
             }
             let filterInput = inputArr[1];
-            filterInput.addEventListener('input', () => {
+            var searchtimer; // timer used to debounce user input into the filter: this reduces redundant back end calls by ensuring the user has paused typing before activating.
+            filterInput.addEventListener('input', () => {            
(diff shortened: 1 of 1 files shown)


mcmonkey4eva:
This PR makes the UI feel significantly less responsive. I really enjoy having the filter list update as soon as I type a couple letters. I want my `offset noise` lora? By the time I've typed `off` there's only a few results, and once I hit `offs` it's only the offset noise one - I can stop and click it well before I've finished typing out `offset noise`.

I've pushed a commit that I think is a bit "best of both worlds": it prevents overlapping multiple calls, but still calls as rapidly as the server is happy to respond.

If this is insufficient for your case, the debounce behavior like this could be added if it's behind a non-default user setting.

brendanhoar:
Saw the other commit first and literally said "oooh!" out loud. 

Thank you for the kind response. I know you did not have this on your agenda for today's coding session.

Once I spend some time with your commit, depending on the results, I may be happy with the results. If not, I may take you up on the offer to consider making the filter search delay optional (perhaps adjustable).

I *think* this category of analysis/resolution will become more important, even for smaller library holders, once mobile use of SwarmUI becomes a hot feature, due to the effects of high latencies/low bandwidths.

B

Issue #19, opened 2024-06-23
Inpainting (with mask shrink grow) & segmenting should use the model native resolution (or current resolution settings) IMHO


CholmesFr:
Hi!
Awesome work with swarmUI.
It's almost perfect for my needs apart for an issue with inpainting and segmenting.

Sometime when using inpainting with **mask shrink grow**, the inpaint is deformed / low quality.
Looking at the generated workflow, it seems that swarm use the current image resolution for the inpaint:
![image](https://github.com/mcmonkeyprojects/SwarmUI/assets/63900078/de212565-bc11-438e-9dd5-caba939b29f0)

Step to reproduce:
1°) import a big image (or scale an image to a big resolution)
2°) inpaint a small part using the **mask shrink grow** option
3°) import the workflow (great option for debuging btw)
4°) check the size of SwarmImageScaleForMP => it's set to the current image resolution

When working in native resolution (eg 1024x1024 for SDXL) it's not an issue.
However when working with a high resolution image, it's problematic because the inpaint resolution can exceed the range of the model and the memory can explode.

Here an example when trying to inpaint a small part of a big image with a sd1.5 model.
After importing it, if I force the native resolution in the workflow we can see that the inpaint lose quality as the resolution grows.
![image](https://github.com/mcmonkeyprojects/SwarmUI/assets/63900078/d76f9473-a8c3-4306-9812-99585b8d9cc8)
(Same seed but Gandy is not recognizable at >1024, the skin is smoother and the glasses frame is deformed)

One solution would be to always use the native resolution in this case: 
WorkflowGenerator.cs line 327 :

string scaledImage = CreateNode("SwarmImageScaleForMP", new JObject()
 {
     ["image"] = new JArray() { croppedImage, 0 },
     ["width"] = UserInput.Get(T2IParamTypes.Model).StandardWidth, // use model native resolution
     ["height"] = UserInput.Get(T2IParamTypes.Model).StandardHeight, 
     ["can_shrink"] = true // Shrink inpaint if mask is too big
 });

It works great but the inpaint is always square (not sure if it's better or not but it should be possible to keep the mask ratio if necessary)

The other solution would be to give the user more control and use the selected resolution in the UI.
I believe some changes in JS part would be necessary as I don't think that these parameters are accessible in the C# function.

Segmenting use the same portion of code but that's great because it suffer the same issue, so both would be fixed.

Let me know what you think and if I should work on a PR.

[...]

PR #52, opened 2024-07-06
fix: use native resolution for segmentation/inpainting ( #19 )

CholmesFr:
As discussed in #19 Segmentation and Inpainting (with Mask Shrink Grow) use the current image resolution to render the masked area.

### This fix ensure that the render is done with the right model resolution depending on the settings : 

- Current model resolution for segmentation without refiner
- Refiner resolution (when enabled) for segmention
- Current model resolution for inpainting with or without refiner as the inpaint is done before the refiner pass

### DIFF
--- a/src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
+++ b/src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
@@ -291,7 +291,7 @@ public class WorkflowGenerator
     /// <param name="vae">The relevant VAE.</param>
     /// <param name="threshold">Optional minimum value threshold.</param>
     /// <returns>(boundsNode, croppedMask, maskedLatent).</returns>
-    public (string, string, string) CreateImageMaskCrop(JArray mask, JArray image, int growBy, JArray vae, double threshold = 0.01, double thresholdMax = 1)
+    public (string, string, string) CreateImageMaskCrop(JArray mask, JArray image, int growBy, JArray vae, double threshold = 0.01, double thresholdMax = 1, bool fromSegmentation = false)
     {
         if (threshold > 0)
         {
@@ -327,9 +327,9 @@ public class WorkflowGenerator
         string scaledImage = CreateNode("SwarmImageScaleForMP", new JObject()
         {
             ["image"] = new JArray() { croppedImage, 0 },
-            ["width"] = UserInput.Get(T2IParamTypes.Width, 1024),
-            ["height"] = UserInput.GetImageHeight(),
-            ["can_shrink"] = false
+            ["width"] = UserInput.GetModelForInpainting(fromSegmentation)?.StandardHeight,
+            ["height"] = UserInput.GetModelForInpainting(fromSegmentation)?.StandardWidth,
+            ["can_shrink"] = true
         });
         string vaeEncoded = CreateVAEEncode(vae, [scaledImage, 0], null, true);
         string masked = CreateNode("SetLatentNoiseMask", new JObject()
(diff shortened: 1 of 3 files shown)


src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
```diff
@@ -371,16 +378,20 @@ public JArray RecompositeCropped(string boundsNode, JArray croppedMask, JArray f
             ["upscale_method"] = "bilinear",
             ["crop"] = "disabled"
         });
-        string thresholded = CreateNode("ThresholdMask", new JObject()
```
mcmonkey4eva, line 374: Feathering doesn't sound bad, but don't get rid of the threshold, that's there very intentionally - removing it will cause obscure issue regressions.

The feathering itself should probably be a separate addition from the rest of this PR being a bugfix, and be controlled by a parameter

src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
```diff
@@ -4,6 +4,7 @@
 using SwarmUI.Text2Image;
 using SwarmUI.Utils;
 using System.IO;
+using static SwarmUI.Utils.PromptRegion;
```
mcmonkey4eva, line 7: accidental stray import

src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
```diff
@@ -326,18 +326,25 @@ public string CreateLoadImageNode(Image img, string param, bool resize, string n
         });
         string scaledImage = CreateNode("SwarmImageScaleForMP", new JObject()
         {
-            ["image"] = new JArray() { croppedImage, 0 },
-            ["width"] = UserInput.Get(T2IParamTypes.Width, 1024),
-            ["height"] = UserInput.GetImageHeight(),
-            ["can_shrink"] = false
+            ["image"] = new JArray() { croppedImage, 0 },
```
mcmonkey4eva, line 329: format

src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
```diff
@@ -326,18 +326,25 @@ public string CreateLoadImageNode(Image img, string param, bool resize, string n
...
             ["samples"] = new JArray() { vaeEncoded, 0 },
             ["mask"] = new JArray() { croppedMask, 0 }
         });
-        return (boundsNode, croppedMask, masked);
+
+        return (boundsNode, croppedMask, skipLatentNoise ? vaeEncoded : latentNoiseMasked);
+    }
+
+    // Segmentation use refiner Model when enabled but inpainting does not
+    private T2IModel GetRelevantModel(bool fromSegmentation)
```
mcmonkey4eva, line 345: This should probably be a sub-method inside of the method it's used by rather than a class method

src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
```diff
@@ -976,6 +977,14 @@ JArray doMaskShrinkApply(WorkflowGenerator g, JArray imgIn)
                 {
                     PromptRegion.Part part = parts[i];
                     string segmentNode;
+                    bool sharper = false;
+
+                    if (part.DataText.StartsWith("sharp-"))
```
mcmonkey4eva, line 982: this is also a full separate new feature addition that should be done separately

PR #55, opened 2024-07-07
[Feature] new option for segmentation : Sharper inpaint 

CholmesFr:
### Option for segmentation : Sharper inpaint 

By default, Segmentation use a 'latent noise mask' node before render to better blend the inpaint part to the original image.
 
 It works great most of the time but as a result the inpainted image can be a little soft.
As an option, I wanted to enable the kind of sharpness you can obtain with Adetailler in Automatic1111.

To use this option, I propose adding a 'sharp-' prefix in the segment parameters:  \<segment:sharp-face\>

When using the '**sharp-**' prefix, we skip the latent noise mask to enable a more raw render which can result on a crisper image (that work very well for face inpainting).

For this to work, I've needed to replace the 'Threshold Mask' by a 'Feather mask'  for a smoother transition.

@mcmonkey4eva : I've tried keeping the Threshold mask before or after the Feather mask but I as a result the mask border is very apparent. I'm not sure why. Anyway, in the code the Feather mask is only enable when using the new option.

A few examples with and without the option (cropped):

image

Remark: This PR is based on PR #52 and should be reviewed after #52 is merged.

### DIFF
--- a/src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
+++ b/src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
@@ -291,7 +291,7 @@ public class WorkflowGenerator
     /// <param name="vae">The relevant VAE.</param>
     /// <param name="threshold">Optional minimum value threshold.</param>
     /// <returns>(boundsNode, croppedMask, maskedLatent).</returns>
-    public (string, string, string) CreateImageMaskCrop(JArray mask, JArray image, int growBy, JArray vae, double threshold = 0.01, double thresholdMax = 1)
+    public (string, string, string) CreateImageMaskCrop(JArray mask, JArray image, int growBy, JArray vae, double threshold = 0.01, double thresholdMax = 1, bool fromSegmentation = false, bool skipLatentNoise = false)
     {
         if (threshold > 0)
         {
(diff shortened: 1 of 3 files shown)


mcmonkey4eva:
Oh, I see my mistake re threshold: I mixed up `SwarmMaskThreshold` (a threshold function node) and `ThresholdMask` (a boolean splitter function).

Checking comfy source, `FeatherMask` does exactly nothing of value here - the improvement you're seeing is just from the wrong threshold node being removed. It's one of the weird core mask nodes that just fiddles the edges of a square mask rather than anything useful with existing masks.

mcmonkey4eva:
The `ThresholdMask` (boolean split) was added because:

There were gray regions of the masks, eg maskval=0.5 or 50%, that would apply the 50% when generating, and then when compositing would do 50% again - effectively an exponential mask decay.

This meant that for partial matches, it would *composite more smoothly* but be basically incapable of content replacement, because the exponential decay would add back too much of the source image.

---

I've added a new parameter as Advanced -> Regional Prompting -> `Mask Composite Unthresholded`, which removes the threshold call when checked.

mcmonkey4eva:
Huh? Bypassing and removing a node are the same thing

mcmonkey4eva:
Ah, so not masking the generation at all, and instead masking the result at the output.
That... seems like it would have more unwanted effects than wanted ones? At least for any non-small creativity value

PR #59, opened 2024-07-10
Fix File Browser Header Path Navigation Issue

maedtb:
Fixed an issue where navigating to a folder by clicking on a folder name in the Header Path navigation (i.e. `Root/Folder/Names/Here/`), would incorrectly add a `t/` folder to the base of the path. We now correctly chop off all of the leading `Root/` prefix when building paths instead of only `Roo`, fixing this issue.


mcmonkey4eva:
Wow, guess I haven't touched those breadcrumb path buttons in a while -- Thank you!

Issue #28, opened 2024-06-26
Feature Request: Progress bar/% amount in tab window


Nitefighter7:
After using auto1111 I swapped to Swarm cause it let me use my GPU for dat speedy iterations.

but when I was using Auto1111 I had a little XX% in the page name, this meant I could have a different tab open and it would tell me how far through the image gen it was.

Similarly an adjacent feature would be some kind of progress bar or estimate in the actual UI, so I can have some idea of when to check back in.

mcmonkey4eva:
In actual UI there's two progress reports:

1st, the queue size estimate
![image](https://github.com/mcmonkeyprojects/SwarmUI/assets/4000772/f21dcf5b-7920-45e8-b529-8dc6f04eb9a4)

2nd, the bars under images:
![image](https://github.com/mcmonkeyprojects/SwarmUI/assets/4000772/6765c772-d63e-4c39-8bad-45c043c81b4b)

(Generally the green bar is the main one, but there's also a purple bar that will show up when you have a custom workflow with many steps to jump through)

mcmonkey4eva:
There's now a `(3 generating)` prefix on the page title for when generating several images

SirVeggie:
Those progress bars should ideally be placed somewhere else, since hiding the image feed also hides the progress bar. This is mainly in consideration of a mobile UI.

PR #87, opened 2024-07-23
Added display of current image progress % to title

crazeejeeves:
Topic: Feature request (#28)

Used the (currently unused) gotProgress callback to translate the overall progress %.

Implementation choices:
1. num_current_gens required to mimic title format within the handler. Opted to pass in as a parameter to not be coupled with the global num_current_gens within the callback
2. Overall vs current? Both seem to reset to 0 per image and not really representative of the batch progress. 'Current' updates more frequently, but overall is closer to the actual progress.

Changing the underlying computation behavior of progress would require refactoring beyond the scope of this tiny feature.

### DIFF
--- a/src/wwwroot/js/genpage/generatehandler.js
+++ b/src/wwwroot/js/genpage/generatehandler.js
@@ -31,8 +31,20 @@ class GenerateHandler {
         return gotImagePreview(image, metadata, batchId);
     }
 
-    gotProgress(current, overall, batchId) {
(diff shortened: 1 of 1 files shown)


mcmonkey4eva:
Going to decline this PR, as it introduces code errors, breaks format, and you'd need to understand the percentages better to properly mix the 3 values (current node, overall workflow, and count of gens in queue)

PR #97, opened 2024-07-28
Fix support for CSS styling alternating image batch thumbnails.

maedtb:
The existing `batchId` approach doesn't work well as it's not a counter. Added a new `numImages` counter on the image container to track how many image have been added to it. This number is reset when a batch is cleared.

### DIFF
--- a/src/wwwroot/js/genpage/main.js
+++ b/src/wwwroot/js/genpage/main.js
@@ -38,7 +38,9 @@ let statusBarElem = getRequiredElementById('top_status_bar');
 
 /** Called when the user clicks the clear batch button. */
 function clearBatch() {
-    getRequiredElementById('current_image_batch').innerHTML = '';
+    let currentImageBatchDiv = getRequiredElementById('current_image_batch')
+    currentImageBatchDiv.innerHTML = '';
+    currentImageBatchDiv.dataset.numImages = 0;
 }
 
 /** Reference to the auto-clear-batch toggle checkbox. */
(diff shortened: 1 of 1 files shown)


src/wwwroot/js/genpage/main.js
```diff
@@ -38,7 +38,9 @@ let statusBarElem = getRequiredElementById('top_status_bar');
 
 /** Called when the user clicks the clear batch button. */
 function clearBatch() {
-    getRequiredElementById('current_image_batch').innerHTML = '';
+    let currentImageBatchDiv = getRequiredElementById('current_image_batch')
```
mcmonkey4eva, line 41: missing `;` on the end

src/wwwroot/js/genpage/main.js
```diff
@@ -723,15 +725,16 @@ function appendImage(container, imageSrc, batchId, textPreview, metadata = '', t
     if (typeof container == 'string') {
         container = getRequiredElementById(container);
     }
-    let div = createDiv(null, `image-block image-block-${type} image-batch-${batchId == "folder" ? "folder" : (batchId % 2)}`);
+    container.dataset.numImages === undefined ? container.dataset.numImages = 1 : container.dataset.numImages++
```
mcmonkey4eva, line 728: missing `;` here too

mcmonkey4eva: also wait wtf is this an ternary statement pair? ow. Use if/else

mcmonkey4eva: _alsoalso_ this variable doesn't need to be in a div dataset, it's just a generic freefloating variable

mcmonkey4eva: Valid point actually, nevermind that comment then

PR #98, opened 2024-07-28
Normalize 'batchId' format.

maedtb:
GridGen and ImageEditBatcher used `batchNumber` and the regular image generator used `batchNumber_batchIndex`.  Since all of these support multiple images per patch, I have normalized this to `batchNumber_batchIndex`.

Also fixed GridGen `GRID_IMAGE` images having the same `batchIndex` as the last image in the batch. These images now have a `batchIndex` one larger.


mcmonkey4eva:
I think that code assumed batch indices would start at 0, not 1, heh.

PR #104, opened 2024-07-28
Improve command line friendliness

maedtb:
I added support for `--key=value` command line arguments (with an equals-sign instead of a space), as well as adding a `--help` option that prints out the basic supported command line options and their values.  The commits themselves have more detail.

### DIFF
--- a/src/Core/Program.cs
+++ b/src/Core/Program.cs
@@ -105,6 +105,11 @@ public class Program
         {
             Logs.Init("Parsing command line...");
             ParseCommandLineArgs(args);
+            if (GetCommandLineFlagAsBool("help", false))
+            {
+                PrintCommandLineHelp();
+                return;
+            }
             Logs.Init("Loading settings file...");
             DataDir = Utilities.CombinePathWithAbsolute(Environment.CurrentDirectory, CommandLineFlags.GetValueOrDefault("data_dir", "Data"));
             SettingsFilePath = CommandLineFlags.GetValueOrDefault("settings_file", "Data/Settings.fds");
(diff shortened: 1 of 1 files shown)


src/Core/Program.cs
```diff
@@ -621,5 +645,88 @@ public static bool GetCommandLineFlagAsBool(string key, bool def)
             var mode => throw new InvalidDataException($"Command line flag '{key}' value of '{mode}' is not valid")
         };
     }
+    public static void PrintCommandLineHelp()
+    {
+        Console.WriteLine("""
+Usage: SwarmUI.dll [option]...
+  or   SwarmUI.dll --help
+
+  Options:
```
mcmonkey4eva, line 654: This should be put in some shared location so it's not duplicated between docs and here.
(Or just link to the docs from here lol)

src/Core/Program.cs
```diff
@@ -621,5 +645,88 @@ public static bool GetCommandLineFlagAsBool(string key, bool def)
             var mode => throw new InvalidDataException($"Command line flag '{key}' value of '{mode}' is not valid")
         };
     }
+    public static void PrintCommandLineHelp()
+    {
+        Console.WriteLine("""
+Usage: SwarmUI.dll [option]...
```
mcmonkey4eva, line 651: running the built .dll directly is not the intended/proper way to run Swarm - `launch-linuxmac.sh` or `launch-windows.bat` should be used (both autoforward args)

mcmonkey4eva:
Is there any reason for the key=value support, other than "That one python arg lib does that, so AI script users are probably used to it"?

PR #106, opened 2024-07-30
Hitting Enter in the Negative Prompt text generates a batch.

maedtb:
This makes the Negative Prompt box behave like the Positive Prompt box.  If you press `Enter`, you will start a generation.  If you press `Shift+Enter` instead, you will get a newline.

### DIFF
--- a/src/wwwroot/js/genpage/main.js
+++ b/src/wwwroot/js/genpage/main.js
@@ -1573,6 +1573,15 @@ function pageSizer() {
             return false;
         }
     });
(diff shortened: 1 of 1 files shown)


aimerib:
This behavior is even worse on mobile, where enter in either box will start generation. On mobile shift+enter is not an intuitive combination. In my PR I turned the enter to gen off for both prompts

mcmonkey4eva:
yeah mobile understandably needs different behavior there than desktop

Issue #39, opened 2024-07-02
Feature Request: Support small form factor devices


hendrix2222:
### Feature Idea

The UI is currently designed for use on desktop devices, and is rendered appropriately for that use-case.

For non-desktop devices such as tablets or other mobile devices, the UI is not rendered in a user-friendly way. Some support for at least basic text-to-image generation on small form factors would be greatly appreciated.

mcmonkey4eva:
Yep, mobile support on the Generate tab is definitely a major feature requirement.

I have a rough idea of how it'd work:

The main screen should be just the prompt box and image result area,
and a button on the left to pull up the parameters list, which pops up to cover the screen and you can edit then swipe it back away
similarly you can pull up the other menus (image history / model browser / etc) and they also temporarily take over the full screen while you use them.

Basically take all the current UI features that are placed into cells, and swap them to being sub-screens to swap between, with the 'main prompt + result' area as the centerpoint

hendrix2222:
Yeah, I'm a backend dev by trade, so UI work isn't something I'm even remotely qualified to speak to, other than it's a lot more difficult than it looks to get things to display nicely across different form factors. And I think you sort of alluded to it, but it may actually end up making the desktop render better by making things a bit more modular anyway.

Separately, I had made a feature request here: https://github.com/ShiftHackZ/Stable-Diffusion-Android/issues/204

to potentially add support for SwarmUI's API to their native Android application, but I have no clue how active the development is there, etc.

jonathan-f-silva:
While this is not properly addressed, I did some (not so) quick hack in the simple tab to at least do some gens on mobile. Here are the changes if anyone is interested:
https://github.com/mcmonkeyprojects/SwarmUI/compare/mcmonkeyprojects:SwarmUI:master...jonathan-f-silva:SwarmUI:feat/simple-mobile?expand=1

<details>
<summary>Demo GIF</summary>

![simple-mobile](https://github.com/mcmonkeyprojects/SwarmUI/assets/85591570/066f913e-2515-49b8-81bf-3f34039189c4)

</details>

mcmonkey4eva:
If you get it clean and it doesn't break on desktop sized usage, I'd love a PR for that!

mcmonkey4eva:
mobile support is in the works, Swarm now has extremely minimal but functional phone screen support. Opens to the center area only, swipe from the edges to pull in the side views.

PR #107, opened 2024-07-30
Add initial mobile support

aimerib:
This change is a bit scary. I'm opening it as a draft because I figured I'd get the code out of my local machine and let people look at it so we could make it nice.

I'm adding your list of desired features from #39 here as a guide to walk towards:

- [x] The main screen should be just the prompt box and image result area
- [ ] A button on the left to pull up the parameters list, which pops up to cover the screen and you can edit then swipe it back away
- [x] You can pull up the other menus (image history / model browser / etc) and they also temporarily take over the full screen while you use them.

Goal: Basically take all the current UI features that are placed into cells, and swap them to being sub-screens to swap between, with the 'main prompt + result' area as the centerpoint

Of the list above:
- I've right now made everything a single column, so there's a lot of scrolling, but I tried to add some logic to jump around to logical points (when in mobile only). I like your idea much better, but that's a much bigger lift. I'm happy to works toward that in this PR before promoting it out of draft.
- Regarding a button on the left, what size of screens are we targeting with that? There's a fine line between target hit area and screen real estate on an iphone 11 for example, so I tried to shy away from anything on the edges, but only because I wasn't creative enough to come up with a clean way to do anything else. A clearly labeled area that can be tapped without interrupting thumb navigation would be really clean!
- So, a little bit like the native fly-out/fly-up pages/screens on native apps?

Things I still need to work on and would like maybe some help/pointers:
- [ ] Size of batch image container is bigger than viewport, even if image is centered on the screen fine. Have to fix the sizing there and remove overflow
- [ ] The FAB button supports long-press for an extra context menu but for some reason I'm not able to replicate that correctly in the browser simulator, but I'll take a screencap of it working on my phone. Just plain weird.
- [ ] The behavior around dragging the top bars is a little wonky right now. All that math was too hard for me at midnight after a long day of parenting. I'll get back to it soon.

Current progres: 

https://github.com/user-attachments/assets/31c30436-c807-48e8-874c-b4e29f5c6c3a


mcmonkey4eva:
`So, a little bit like the native fly-out/fly-up pages/screens on native apps?` yes

` Size of batch image container is bigger than viewport, even if image is centered on the screen fine. Have to fix the sizing there and remove overflow` on mobile the batch view should probably be moved to the bottom or something instead of the side, at least when the screen is vertical

` The FAB button supports long-press for an extra context menu but for some reason I'm not able to replicate that correctly in the browser simulator, but I'll take a screencap of it working on my phone. Just plain weird.` The what does what

` The behavior around dragging the top bars is a little wonky right now. All that math was too hard for me at midnight after a long day of parenting. I'll get back to it soon.`
on mobile the bar/dragging thing should be removed entirely in favor of popup screens

aimerib:
`"Size of batch image container is bigger than viewport, even if image is centered on the screen fine. Have to fix the sizing there and remove overflow" on mobile the batch view should probably be moved to the bottom or something instead of the side, at least when the screen is vertical` right. It is actually right under the main current_image containter. But each batch image also has their own container with a purple border denoting active batch image. That container is currently wider than the viewport, allowing users to scroll sideways. That is unintended behavior, and I need to fix that container sizing.

 `"The FAB button supports long-press for an extra context menu but for some reason I'm not able to replicate that correctly in the browser simulator, but I'll take a screencap of it working on my phone. Just plain weird." The what does what` FAB = Floating Action Button. sorry, I was borrowing lingo from Material Design here. It's that big circle button floating on mobile UIs, usually on the bottom corner, usually with a plus icon. It usually expands out into more buttons or activates the main UI functionality. In the video it is the big purple button on the lower right corner.

`"The behavior around dragging the top bars is a little wonky right now. All that math was too hard for me at midnight after a long day of parenting. I'll get back to it soon."
on mobile the bar/dragging thing should be removed entirely in favor of popup screens` Right. They are. This is for the desktop version that I am talking about. I had to move the settings around so I could style things correctly without reaching for JS. But since we are talking about adding flyouts and such, I think we won't be able to escape using JS for moving things around anyways. It's a little bit of a catch 22 here, in that the perfect version of the mobile UI will take a lot more effort and time to finish but would allow 0 changes to the current layout of desktop. I wouldn't call moving the order of the layout columns a small change necessarily, but I figured I'd showcase the changes here to see if they are even acceptable. That would allow official mobile support within weeks versus months, but understandably not a change to be taken lightly

AshtakaOOf:
This PR is quite far behind the master branch.
Would it be possible to update it?

mcmonkey4eva:
Closing for now as continuing this would require basically starting over from how long it's been since updating

PR #109, opened 2024-07-30
Add View Grid option to imagehistorybrowser-header

aimerib:
I couldn't find a way to get back to the grid html after I navigated away from the Grid Generation tool view, so I tried to make it a little easier to get to that when I'm not sitting at my main computer with access to the file browser.
Screenshot 2024-07-30 at 10 12 45 AM

### DIFF
--- a/src/wwwroot/js/genpage/browsers.js
+++ b/src/wwwroot/js/genpage/browsers.js
@@ -180,6 +180,18 @@ class GenPageBrowserClass {
         let rootPathPrefix = 'Root/';
         let partial = '';
         for (let part of (rootPathPrefix + path).split('/')) {
+            if (part.includes('grid-')) {
+                let gridLink = document.createElement('a');
+                gridLink.href = `/View/local/${path}/index.html`;
+                gridLink.target = '_blank';
+                gridLink.className = 'grid-view-link';
+                gridLink.textContent = 'View Grid';
+                // I wanted the view grid link to come before the path area so using prepend here.
+                // happy to change this if needed.
+                pathGen.prepend(document.createTextNode(' • '));
+                pathGen.prepend(gridLink);
+                pathGen.prepend(document.createTextNode('  '));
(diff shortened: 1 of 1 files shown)


aimerib:
One concern I have is that not every Grid is generated as Web in Gridgen. I might need to add some way to distinguish if this specific grid that we are looking at has an html page or not. Maybe this was too hastily cooked, but I would love ideas if any. I can also close this until I can think of a better way.

mcmonkey4eva:
you can just click on the `index.html`?

aimerib:
If I am sitting at the same computer that has the actual files, sure. But if I am accessing the UI from my phone, or if the page was refreshed, I no longer have access to the full path of the grid. it would be something like: `http://192.168.0.112:7862/View/local/Grids/grid-2024-07-30-02-45-41/gridanemhere/index.html` which it is pretty easy to navigate to or find from the computer that has the files.

But once you navigate away from the gridgen screen, I no longer have the link to the grid I started available. It uses datetime as a path, which means if I started the grid at 7:34:21pm I have to remember that exact time if I have hopes to get back to it. Right now my process is: start the grid gen, text the full url to myself (local network ip) and then walk away. After an hour or so I check the grid link and if the gen finished all the pictures will be there.

mcmonkey4eva:
image

aimerib:
🤦🏻‍♂️  my grids are so massive I didn't see the html right there in front of me. I even took a screenshot of it. damn...

aimerib:
Real question then: Wouldn't it make sense for the .html link to be the first item to be showed in this view, regardless of sorting? (actual question, not a suggestion or request for change)

mcmonkey4eva:
yes

mcmonkey4eva:
And now they do

aimerib:
Thank you so much! I just tested it and it is immediately useful! 🙏🏻

Issue #92, opened 2024-07-27
Use dimensions of selected image when doing img2img


Dwedit:
### Feature Idea

If you are doing an img2img, you probably want to output to be the size of the image you selected rather than the default size.  I'd request a checkbox for 'use selected image dimensions'.

Yes, you can currently manually type in the image dimensions somewhere else on the page.  It's a pain to do that.

Dwedit:
A solution to this issue would require not only providing the coordinates, but also adding support for coordinates that are not multiples of 8.

There is a comfy UI extension called [comfyui-image-round](https://github.com/cdb-boop/comfyui-image-round) which provides a Padding node and a Cropping node.  The Padding node pads the image to a multiple of 8 (or any other desired number), and the Cropping node will crop the image.  You can also connect the Cropping node to the Padding node, and it will undo the padding added to the image.

When the base image is already a multiple of 8, no cropping or padding is required.  But if it's not, you need padding and cropping to match the exact resolution of the input image.  Using comfy nodes to do this is the most direct way.

A potential problem is that the padding node wants to pad with a single solid color.  I don't know if that is the best idea or not.  An alternative could be to pad the image with a flipped version of the image.  I don't know which will produce less artifacts when it goes through a VAE encoder.

mcmonkey4eva:
added, button next to Init Image input

PR #133, opened 2024-08-05
Feature to use dimensions of the Init Image when doing Img2img

Dwedit:
For Issue #92

This adds a "Use Dimensions of Init Image" check box after the "Init Image" field.

When checked, it will use the dimensions of the Init Image instead of the selected resolution *unless* a size is specified in "Swarm Internal -> Raw Resolution", then it will pick that.

### DIFF
--- a/src/Text2Image/T2IParamInput.cs
+++ b/src/Text2Image/T2IParamInput.cs
@@ -30,6 +30,17 @@ public class T2IParamInput
         },
         input =>
         {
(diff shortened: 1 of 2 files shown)


Dwedit:
I wonder if I should cancel this PR and also add in support for using the dimensions of ControlNet images.

mcmonkey4eva:
this should be done via the UI rather than the server - when an image is added, swap the current resolution to match t he image, based on a user setting

PR #155, opened 2024-08-09
Add model detection for Stable Cascade models

maedtb:
This seemed to work correctly with all my cascade models and had no false-positives when I rebuilt my metadata.


mcmonkey4eva:
who the heck is publishing cascade models without metadata, argh

PR #230, opened 2024-08-31
Save Models in baseModel subfolder

aimerib:
Add the functionality of reading baseModel from civitai metadata, and save models in their base model subfolders. This will look like: Models/Stable-diffusion/
  SDXL 1.0/
  SD1.5/
  Pony/
  Flux.1 D/
  Flux.1 F/
etc...

This is exactly the same changes as shared in Discord. I welcome any feedback and I'm happy to make any changes necessary. This is just my own little patch that I apply after any updates because I like my models to stay organized in basemodel subfolders since it helps identify them at a glance in the UI.

### DIFF
--- a/src/Core/Settings.cs
+++ b/src/Core/Settings.cs
@@ -273,6 +273,9 @@ public class Settings : AutoConfiguration
         [ConfigComment("If true, folders will be discarded from starred image paths.")]
         public bool StarNoFolders = false;
 
+        [ConfigComment("Whether to automatically use the base model type as part of the model path when downloading using the 'Model Download' tool")]
+        public bool GroupDownloadedModelsByBaseType = false;
+
         public class ThemesImpl : SettingsOptionsAttribute.AbstractImpl
         {
             public override string[] GetOptions => [.. Program.Web.RegisteredThemes.Keys];
--- a/src/WebAPI/ModelsAPI.cs
+++ b/src/WebAPI/ModelsAPI.cs
@@ -505,7 +505,21 @@ public static class ModelsAPI
         }
         try
         {
-            string outPath = $"{handler.FolderPaths[0]}/{name}.safetensors";
+            string baseModel = "";
+            if (!string.IsNullOrWhiteSpace(metadata))
+            {
+                JObject metadataObj = JObject.Parse(metadata);
+                baseModel = metadataObj["modelspec.baseModel"]?.ToString() ?? "";
+            }
+            string outPath;
+            if (!string.IsNullOrWhiteSpace(baseModel))
+            {
+                outPath = $"{handler.FolderPaths[0]}/{baseModel}/{name}.safetensors";
+            }
+            else
+            {
+                outPath = $"{handler.FolderPaths[0]}/{name}.safetensors";
+            }
             if (File.Exists(outPath))
             {
                 await ws.SendJson(new JObject() { ["error"] = "Model at that save path already exists." }, API.WebsocketTimeout);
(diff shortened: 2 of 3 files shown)


aimerib:
@maedtb actually really good points. I pushed another commit that makes this a user setting default to off. That way it doesn't break existing behavior, and allows users to opt-in to using this option. Would that, plus using `Utilities.StrictFilenameClean()` be a sufficient approach? I am hesitant to add some behavior to the UI that users might want to disable by default, therefore just annoying them.

mcmonkey4eva:
please use regular commits btw, don't force push

aimerib:
Ah, sorry. I'm used to how my workplace does things. We only ever have a max of 3 commits per PR, so everything ends up with a `git commit --amend` and a `git push --force-with-lease` to keep the commits that show in the main repo clean. I will refrain from doing so in the future

mcmonkey4eva:
Oughtta tell them that github has a button to squash the PR after

image

So you can make PRs readable and diffs decipherable, _and also_ keep the main repo commit history clean

src/Core/Settings.cs
```diff
@@ -273,6 +273,9 @@ public class FileFormatData : AutoConfiguration
         [ConfigComment("If true, folders will be discarded from starred image paths.")]
         public bool StarNoFolders = false;
 
+        [ConfigComment("Whether to automatically use the base model type as part of the model path when downloading using the 'Model Download' tool")]
```
mcmonkey4eva, line 276: 3x minor doc nits:
- `.` at end of the sentence
- also it's the `Model Downloader utility`
- Also doc here should ideally be a bit more clean what it does - something like `\nFor example, SDXL models will be placed automatically into an `SDXL/` directory.

aimerib:
We've talked about doing that. We have to be SOC2 and ISO<long number> compliant, so it is about having processes and following them, not about the processes making any sense. Changing any process is a huddle because it requires a security committee meeting and approval, and it has to then get documented. It's a lot of hassle for what is easier to just adapt lol

src/Core/Settings.cs
aimerib: On it

src/WebAPI/ModelsAPI.cs
```diff
@@ -505,8 +505,23 @@ public static async Task<JObject> DoModelDownloadWS(Session session, WebSocket w
...
+            string baseModel = "";
+            if (!string.IsNullOrWhiteSpace(metadata))
+            {
+                JObject metadataObj = JObject.Parse(metadata);
+                baseModel = metadataObj["modelspec.baseModel"]?.ToString() ?? "";
+            }
+            string modelOutPath;
+            if (!string.IsNullOrWhiteSpace(baseModel) && session.User.Settings.GroupDownloadedModelsByBaseType)
+            {
+                modelOutPath = $"{handler.FolderPaths[0]}/{baseModel}/{name}.safetensors";
```
mcmonkey4eva, line 517: Since this is just editing the folder path, it should be implemented in the `utiltab.js` rather than the C# here.
On top of avoiding cleaning/handling/API stability issues, that also gives the perk that you can just emit the folder prefix into the existing folder dropdown field, ie allowing users to edit the path on a case-by-case basis whenever they want

aimerib: How do I get access to user settings from the js side? On the C# side I could just reach for the user session pretty easily. I know the JS side must have a helper method for that as well, right?

aimerib: `src/wwwroot/js/genpage/settings_editor.js` `getUserSetting` maybe?

mcmonkey4eva: yeah `getUserSetting`

mcmonkey4eva:
Closing this PR as unmaintained, but also the model downloader should probably have general syntax for path building rather than a dedicated baseModel input, other requests for options to autocontrol paths have been requested

PR #238, opened 2024-09-02
Add per-user configurable top-level tab hiding

neggles:
hi fren, it's your worst nightmare, I actually did the thing! and it even accounts for dynamic/extension-added tabs!

well okay it only sort of works. by which I mean it does not work.

the magicks for dropdown popovers appear to be broken with select2's multi-select (since nothing else seems to ever actually call `makeMultiselectInput()`?) and I am not enough of a JS person to work out how to fix that. So the multiselect shows up, but it's styled wrong, and changing the selected options doesn't mark the option as altered so it's not saveable.

But the hiding _does_ work if I manipulate things in the browser debugger to pretend like the server returned a list of tab IDs to hide.

what do, how fix?


src/Core/WebServer.cs
```diff
@@ -13,6 +13,7 @@
 using SwarmUI.WebAPI;
 using System.IO;
 using System.Net;
+using System.Text.RegularExpressions;
```
mcmonkey4eva, line 16: (nit) accidental import there

src/Core/WebServer.cs
```diff
@@ -63,6 +64,31 @@ public static void SetHost(string host, int port)
     /// <summary>Extra content for the Text2Image page's tab bodies. Automatically set based on extensions.</summary>
     public static HtmlString T2ITabBody = new("");
 
+    /// <summary>Set of registered top-level nav tabs.</summary>
+    public Dictionary<string, TabData> RegisteredTabs = new() {
```
mcmonkey4eva, line 68: (nit) the `{` goes on its own line

src/Core/WebServer.cs
```diff
@@ -63,6 +64,31 @@ public static void SetHost(string host, int port)
...
+    public record class TabData(string ID, string Name) { }
+
+    /// <summary>Register a top-level nav tab.</summary>
+    public void RegisterTab(TabData tab)
+    {
+        RegisteredTabs.Add(tab.ID, tab);
+    }
+
+    /// <summary>Register a top-level nav tab from an extension.</summary>
+    /// <param name="ID">The tab button's <a> element ID</param>
```
mcmonkey4eva, line 85: (nit) `.` at end

src/WebAPI/BasicAPIFeatures.cs
```diff
@@ -514,6 +515,22 @@ public static async Task<JObject> SetParamEdits(Session session, JObject rawData
...
+        JArray rawTabs = (JArray)rawData["tabs"];
+        foreach (JObject rawTab in rawTabs.FilterCast<JObject>())
+        {
+            WebServer.TabData tab = new(rawTab["id"].ToString(), rawTab["name"].ToString());
+            if (!Program.Web.RegisteredTabs.ContainsKey(tab.ID))
+            {
+                Logs.Debug($"User '{session.User.UserID}' registered new tab '{tab.Name}' with ID '{tab.ID}'.");
+                Program.Web.RegisterTab(tab);
+            }
+
```
mcmonkey4eva, line 529: (nit) stray blank line

src/WebAPI/BasicAPIFeatures.cs
```diff
@@ -514,6 +515,22 @@ public static async Task<JObject> SetParamEdits(Session session, JObject rawData
         return new JObject() { ["success"] = true };
     }
 
+    public static async Task<JObject> RegisterTabs(Session session, JObject rawData)
```
mcmonkey4eva, line 518: the heck are clients registering tabs for? That's... backwards. Server should define them.

mcmonkey4eva:
Multi-select inputs work as intended locally:
image

Scorers and LoRA params are the only multiselects I think exist offhand. Definitely no settings that use em atm, so might need to adjust some code related to Settings to make it process those properly. (Or be lazy and just do a comma separated list as a text input as I did for reuse params exclude list)

src/Core/WebServer.cs
neggles: dangit
i had a regex here at one point but i got rid of it and forgot this, will fix

src/WebAPI/BasicAPIFeatures.cs
neggles: Because of this:
https://github.com/mcmonkeyprojects/SwarmUI/blob/d611e91b7df2eb1c91e7213367bf7c89430195db/src/Pages/Text2Image.cshtml#L27-L31

unless i am mistaken extensions can add arbitrary extra `<li>` elements (or any html actually) to the tab list via this, and there's no server side registry of what's added - that variable is just An String. Enforcing that extensions provide some kind of ExtensionTab record rather than An String would be an API break and I ain't about to go parsing arbitrary extension-provided HTML in server side code

if i *am* mistaken/didn't look hard enough and this html is generated somewhere from an object rather than being arbitrary shit provided by an estension that would be great

mcmonkey4eva: Yes that's entirely generated, `WebServer#GatherExtensionPageAdditions` generates that, and knows exactly what titles/ids are in use.

neggles: well that'll teach me to go modifying an unfamiliar codebase at 1:30am. 

ok lemme just. go right ahead and make this less insane

mcmonkey4eva:
please don't force push

neggles:
@mcmonkey4eva ok, 2 more force-pushes to undo the first one sry about that 😅

src/Core/Settings.cs
```diff
@@ -284,6 +284,17 @@ public class ThemesImpl : SettingsOptionsAttribute.AbstractImpl
         [SettingsOptions(Impl = typeof(ThemesImpl))]
         public string Theme = "modern_dark";
 
+        public class HiddenTabsImpl : SettingsOptionsAttribute.AbstractImpl
+        {
+            public override string[] GetOptions => [.. Program.Web.RegisteredTabs.Keys];
+
+            public override string[] Names => Program.Web.RegisteredTabs.Values.Select(v => v.Name).ToArray();
```
mcmonkey4eva, line 291: I believe Visual Studio will be highlighting this with a notice to swap it to `[.. ]` syntax?

neggles: I was just copying from `RegisteredThemes`;

https://github.com/mcmonkeyprojects/SwarmUI/blob/3fba08da2fa766c80690a8da10cedcddb4df617f/src/Core/Settings.cs#L276-L281

VSCode .net extn's did not, which is weird since they did for a lot of other stuff. Asked copilot and changed it tho.

neggles:
OK, made that change and un-drafted this since it's not nearly as cursed now.

src/Core/WebServer.cs
```diff
@@ -63,6 +63,25 @@ public static void SetHost(string host, int port)
...
+    public record class TabData(string ID, string Name) { }
+
+    /// <summary>Register a top-level nav tab.</summary>
+    public void RegisterTab(TabData tab)
+    {
+        RegisteredTabs.Add(tab.ID, tab);
+    }
+
+    /// <summary>Register a top-level nav tab from an extension.</summary>
+    /// <param name="ID">The tab button's <a> element ID.</param>
```
mcmonkey4eva, line 78: I think this `<a>` will yield a C# xml doc error?

mcmonkey4eva:
per Discord discussion yesterday: should swap this down to just a CSV input for now.

I posted an issue https://github.com/mcmonkeyprojects/SwarmUI/issues/258 to track the need to replace select2 with something that doesn't suck

neggles:
Still haven't gotten to this but will attempt to in the near future 😅

mcmonkey4eva:
Closing as broadly outdated

Issue #167, opened 2024-08-15
Wildcard Variables to callback previous wildcard values


Targren:
### Feature Idea

When using wildcards in a prompt, users may want to be able to refer to the same value multiple times in the prompt. Dynamic Prompts for A1111 [supports this with its variable syntax](https://github.com/adieyal/sd-dynamic-prompts/blob/main/docs/SYNTAX.md#variables)]. 

e.g.  prompts like these (using a hypothetical syntax to assign a variable)

`an idyllic image of an average family outing on a <random:spring,summer,autumn,winter:var_season> afternoon, in a nature park. The family is wearing casual <var:var_season> weekend clothes. The family dog is chasing a ball while it plays with the children. `

or 

`a mature woman born in 1980, looking at viewer, <wildcard:facial_expression:var_expr>, hazel eyes, dark hair, 
, biker tattoos, leather vest, jeans, leaning on a motorcycle
<segment:face>hazel eyes, laugh lines, <var:var_expr>`

 (Not good prompts, just throw-outs to illustrate some use cases)

mcmonkey4eva:
added by https://github.com/mcmonkeyprojects/SwarmUI/pull/242

PR #241, opened 2024-09-03
Add "variable" storage to wildcard and random tags, to callback random values later in the prompt. 

Targren:
#167 

Append "$$var_name" to "wildcard" or "random" tags to assign, and callback with <var:var_name>


mcmonkey4eva:
This syntax hurts a bit.

It should probably be more like... `<setvar[var_name]:<wildcard:x>> yadda yadda <var:var_name>`

Targren:
Good idea! Then it won't just be limited to those two tags, they can use it for whatever.

PR #255, opened 2024-09-10
Add 16bits color depth support

aimerib:
This PR came up from a conversation on discord with burnyjax regarding saving images with a 16bits color depth. Generally speaking, its pretty hard to distinguish visually between 8bits and 16bits, so for display 8bits is sufficient. The use-case for 16bits is usually in photo-editing.

A specific scenario is product photography. When I use image2image to create a scene around a product, that is just the first step in the process, and later processing is usually a part of the workflow, where having access to the full depth of colors of 16bits makes a huge difference, and helps avoid color-banding and other visual artifacts.

Pillow doesn't support natively so OpenCV was used when processing images requested as 16bits. That image is later converted to a Pillow image when passing it to the comfy ProgressBar. When requesting an image as 8bits RGB (the default for all formats), or when requesting a format other than PNG, the original image processing is used.

burnyjax might be able to give even more specific context for how this might be useful. 

This change will add a boolean flag under Swarm Internals group, and when checked it will save the output image as a 16bits PNG (if png was selected, otherwise this flag will have no effect)


aimerib:
I used this script to verify that the resulting images were actually 16bits and not just 8bits due to converting it to PIL later:
```python
import sys
import struct

def get_png_bit_depth(file_path):
    with open(file_path, 'rb') as f:
        # Read the first 8 bytes to check PNG signature
        signature = f.read(8)
        if signature != b'\x89PNG\r\n\x1a\n':
            raise ValueError("Not a valid PNG file")

        # Read chunks
        while True:
            # Read length of chunk
            length_data = f.read(4)
            if len(length_data) < 4:
                raise ValueError("Unexpected end of file")
            length = struct.unpack('>I', length_data)[0]

            # Read chunk type
            chunk_type = f.read(4)
            if chunk_type == b'IHDR':
                # Read IHDR data
                ihdr_data = f.read(length)
                # Extract bit depth (byte 8 of IHDR)
                bit_depth = ihdr_data[8]
                if bit_depth == 8:
                    return "8-bit"
                elif bit_depth == 16:
                    return "16-bit"
                else:
                    return f"Unsupported bit depth: {bit_depth}"

            # Skip other chunks
            f.seek(length + 4, 1)  # Skip chunk data and CRC

            # If we reach IEND, the file ends
            if chunk_type == b'IEND':
                break

    return "Bit depth not found"

if __name__ == "__main__":
    if len(sys.argv) != 2:
        print("Usage: python check_bit_depth.py <image_file>")
        sys.exit(1)

    file_path = sys.argv[1]
    try:
        bit_depth = get_png_bit_depth(file_path)
        print(f"Bit depth: {bit_depth}")
    except Exception as e:
        print(f"Error: {e}")
        sys.exit(1)
```
running it like: `python bit_depth_checker.py /path/to/gen/image.png`

Not a requirement, just an extra QA step to verify that things still worked as expected.

Screenshot 2024-09-10 at 2 02 57 AM

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmSaveImageWS.py
```diff
@@ -12,7 +13,8 @@ class SwarmSaveImageWS:
     def INPUT_TYPES(s):
         return {
             "required": {
-                "images": ("IMAGE", )
+                "images": ("IMAGE", ),
+                "bit_depth": (["8", "16"], {"default": "8"})
```
mcmonkey4eva, line 17: needs to be under `optional` as its a new input (otherwise it's a breaking change on presaved workflows)

src/Text2Image/T2IParamTypes.cs
```diff
@@ -575,6 +575,9 @@ static List<string> listVaes(Session s)
                 return s;
             }
             ));
+        BitDepth = Register<bool>(new("16Bit Depth", "Optional override for the bit depth of the image.\n If enabled, the image will be saved as a 16-bit PNG.\nIf disabled, the image will be saved as an 8-bit PNG.\n8 is the default and recommended value for most images.\n16 is only recommended for very high quality or print output.\nIf you are not using PNG format, this option is ignored.",
```
mcmonkey4eva, line 578: probably should be a dropdown with 8/16, same as the python code has

src/Text2Image/T2IParamTypes.cs
```diff
@@ -575,6 +575,9 @@ static List<string> listVaes(Session s)
...
+            "false", IgnoreIf: "false", IsAdvanced: true, Group: GroupSwarmInternal, AlwaysRetain: true, OrderPriority: 3
```
mcmonkey4eva, line 579: `AlwaysRetain: true` shouldn't be set

src/Utils/Image.cs
```diff
@@ -31,6 +32,9 @@ public enum ImageType
     /// <summary>File extension for this image.</summary>
     public string Extension;
 
+    /// <summary>The bit depth of the image data.</summary>
+    public int BitDepth { get; private set; }
```
mcmonkey4eva, line 36: get rid of that `{ get; private set; }` mess

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmSaveImageWS.py
```diff
@@ -22,19 +24,46 @@ def INPUT_TYPES(s):
...
 
-    def save_images(self, images):
+    def save_images(self, images, bit_depth):
         pbar = comfy.utils.ProgressBar(SPECIAL_ID)
         step = 0
         for image in images:
-            i = 255.0 * image.cpu().numpy()
-            img = Image.fromarray(np.clip(i, 0, 255).astype(np.uint8))
+            if bit_depth == "16":
+                img_np = image.cpu().numpy()
```
mcmonkey4eva, line 32: this line is duplicated on both sides of the if/else

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmSaveImageWS.py
```diff
@@ -22,19 +24,46 @@ def INPUT_TYPES(s):
...
+
+                if img_np.ndim != 3 or img_np.shape[2] != 3:
```
mcmonkey4eva, line 34: why

aimerib: That checks if the ndim (number of dimensions in an image: height, width, and color channels) is not three (not a standardad RGB image) or if the shape of the third dimension (color channels) is not three (not an RGB).
This prevents the erroneous processing of things like:
Grayscale images (2D arrays)
Images with alpha channels (4 channels instead of 3)
Any other non-standard image format

Considering that we will be receiving this from comfyui, we could assume that we will always receive good data, and forego this check. A grayscale image generated by any checkpoint would still be 8bit, so this is might be redundant. I'm happy to take this out if you also think so. The reason this came to be was that I was testing this script locally first, and the images that I had picked to test turned out to be buggy (synthetically generated by another application for a different purpose). This helped me suss that out and get back on the right path. Different situations there.

mcmonkey4eva: But why is this check needed for 16bit depth, but not 8bit? If the 8bit code handles nonstandard image formats, 16bit should too. if the 8bit code can't, the check should be outside the if block.

aimerib: Really good point. Didn't think of that. Will implement!

mcmonkey4eva:
Tiff addition seems ... fully separate to the rest of this PR, and shouldn't be here.

src/Text2Image/T2IParamTypes.cs
```diff
@@ -575,8 +575,8 @@ static List<string> listVaes(Session s)
                 return s;
             }
             ));
-        BitDepth = Register<bool>(new("16Bit Depth", "Optional override for the bit depth of the image.\n If enabled, the image will be saved as a 16-bit PNG.\nIf disabled, the image will be saved as an 8-bit PNG.\n8 is the default and recommended value for most images.\n16 is only recommended for very high quality or print output.\nIf you are not using PNG format, this option is ignored.",
-            "false", IgnoreIf: "false", IsAdvanced: true, Group: GroupSwarmInternal, AlwaysRetain: true, OrderPriority: 3
+        BitDepth = Register<string>(new("Color Depth", "Specifies the color depth for PNG and TIFF formats:\n- PNG supports 8-bit and 16-bit\n- TIFF supports 8-bit, 16-bit, and 32-bit\n- Other formats ignore this setting\n\nIf 32-bit is selected for PNG, it will default to 8-bit.",
+            "8bit", GetValues: (_) => ["8bit///8-bit (256 colors)", "16bit///16-bit (65,536 colors)", "32bit///32-bit (16.7 million colors)"], IsAdvanced: true, Toggleable: true, Group: GroupSwarmInternal, OrderPriority: 3
```
mcmonkey4eva, line 579: These descriptions are wrong -- "8 bit color" in this context is actually 24bit color for mathematical purposes (8 times 3 channels), there's far more than 256 colors. The description should probably read more like `8-bit per channel (24 bit total)` rather than trying to define the color count (as it's a silly high number in all cases)

aimerib:
I do agree that it seems a little disconnected from the whole thing. Maybe there's a different path here.

A little of the reasoning behind this: Since I was going to change the ParamType from a bool to a list, and burnyjax asked if 32bit support would also be possible, it seemed an easy addition. I didn't realize at the time that the more common format that supports true 32bit (128bpp) colors is tiff. Adding tiff support seemed like a natural progression of this work, since that's not really that common in other formats, as far as I am familiar with.

I could extract the 32bit + tiff support into a separate PR to be merged at some later time after this lands, that way we can keep things separate. 
Tiff support involves a new library for the backend (for 32bit support since ImageSharp doesn't support encoding to 32bit natively). It also involves a new library for the browser to help display tiff in the browser, otherwise generating a tiff image would break preview and display of the image history.

Let me know if you want me to extract the tiff + 32bit support from this PR into a separate one, leave as is, or just remove it altogether (maybe supporting tiff is not that interesting for the project for example)

aimerib:
Valid point. I've extracted the 32bit and tiff parts out (as they are only really relevant together), and this PR can remain focused on supporting 16bits for PNG. I've also addressed your feedback about the description

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmSaveImageWS.py
```diff
@@ -22,19 +26,48 @@ def INPUT_TYPES(s):
...
-    def save_images(self, images):
+    def save_images(self, images, bit_depth):
         pbar = comfy.utils.ProgressBar(SPECIAL_ID)
         step = 0
         for image in images:
-            i = 255.0 * image.cpu().numpy()
-            img = Image.fromarray(np.clip(i, 0, 255).astype(np.uint8))
+            img_np = image.cpu().numpy()
+
+            if img_np.ndim != 3 or img_np.shape[2] != 3:
```
mcmonkey4eva, line 35: earlier comment still stands on this

src/Text2Image/T2IParamTypes.cs
```diff
@@ -91,7 +91,7 @@ public enum ParamViewType
 /// <param name="Subtype">The sub-type of the type - for models, this might be eg "Stable-Diffusion".</param>
 /// <param name="ID">The raw ID of this parameter (will be set when registering).</param>
 /// <param name="SharpType">The C# datatype.</param>
-///
```
mcmonkey4eva, line 94: what

aimerib: I think there was an empty space after the `///` and my editor must have formatted that out

src/wwwroot/js/genpage/browsers.js
```diff
@@ -549,7 +549,7 @@ class GenPageBrowserClass {
             if (!this.showDisplayFormat) {
                 formatSelector.style.display = 'none';
             }
-            let buttons = createSpan(`${this.id}-button-container`, 'browser-header-buttons',
```
mcmonkey4eva, line 552: what

src/wwwroot/js/genpage/main.js
```diff
@@ -455,7 +455,7 @@ function shiftToNextImagePreview(next = true, expand = false) {
 
 window.addEventListener('keydown', function(kbevent) {
     let isFullView = imageFullView.isOpen();
-    let isCurImgFocused = document.activeElement &&
```
mcmonkey4eva, line 458: what

aimerib: same as the other. Empty space at the end, after `&&`

src/wwwroot/js/genpage/browsers.js
aimerib: empty space at the end of the line. my editor autoformatted that. I had earlier code implementation here that I removed, and must not have noticed all the empty line removal

src/Accounts/Session.cs
```diff
@@ -161,7 +161,9 @@ public void Dispose()
         }
         string metadata = user_input.GenRawMetadata();
         string format = user_input.Get(T2IParamTypes.ImageFormat, User.Settings.FileFormat.ImageFormat);
-        image = image.ConvertTo(format, User.Settings.FileFormat.SaveMetadata ? metadata : null, User.Settings.FileFormat.DPI);
+        string selectedBitDepth = user_input.Get(T2IParamTypes.BitDepth, "8bit");
```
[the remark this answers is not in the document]
aimerib: Really good point. I think I just wasn't seeing the forest for the trees. The backend already did all the work for us, no point in doing it all over again. I did have to add a check on the returned image to check how many bits it was encoded with because ImageSharp will automatically encode the image with 8bits if we don't provide the bit depth variant.

src/wwwroot/js/genpage/browsers.js
aimerib: In vscode, trailing whitespaces are by default removed. To prevent that behavior this setting is needed: `files.trimTrailingWhitespace: false`
I went ahead and reverted removing those whitespaces accidentally.

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmSaveImageWS.py
```diff
@@ -22,19 +26,44 @@ def INPUT_TYPES(s):
...
         step = 0
         for image in images:
-            i = 255.0 * image.cpu().numpy()
-            img = Image.fromarray(np.clip(i, 0, 255).astype(np.uint8))
+            img_np = image.cpu().numpy()
+            if bit_depth == "16bit":
+                img_np = np.clip(img_np * 65535.0, 0, 65535).astype(np.uint16)
+                img = self.convert_opencv_to_pil(img_np)
+            else:
+                i = np.clip(img_np * 255.0, 0, 255).astype(np.uint8)
```
mcmonkey4eva, line 38: I again must ask, why was this code altered?

src/Utils/Image.cs
```diff
@@ -280,7 +281,11 @@ public Image ConvertTo(string format, string metadata = null, int dpi = 0)
         switch (format)
         {
             case "PNG":
-                PngEncoder encoder = new() { TextCompressionThreshold = int.MaxValue };
+                PngEncoder encoder = new()
+                {
+                    TextCompressionThreshold = int.MaxValue,
+                    BitDepth = img.PixelType.BitsPerPixel == 8 ? PngBitDepth.Bit8 : PngBitDepth.Bit16
```
mcmonkey4eva, line 287: Should swap this check - `== 16 ? Bit16 : Bit8` - ie for any unexpected value, default to 8

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmSaveImageWS.py
aimerib: I don't think I understand the question.
Why did I write this code, as in what is it doing now? Or why is it written this specific way, rather than

```
if bit_depth == "16bit":
    img_np = np.clip(img_np * 65535.0, 0, 65535).astype(np.uint16)
    img = self.convert_opencv_to_pil(img_np)
else:
    i = 255.0 * image.cpu().numpy()
    img = Image.fromarray(np.clip(i, 0, 255).astype(np.uint8))
```

If the first, the code was modified to handle the expanded value range (0-65535) and data type (uint16)

If the second, maybe just not as much elegance in writing pythonic code.

```
img_np = image.cpu().numpy()
if bit_depth == "16bit":
    img_np = np.clip(img_np * 65535.0, 0, 65535).astype(np.uint16)
    img = self.convert_opencv_to_pil(img_np)
else:
    img = Image.fromarray(np.clip(img_np * 255.0, 0, 255).astype(np.uint8))
```

If the critique is on something else I'm not seeing, I'd appreciate the pointer on what is it that I'm missing

src/Utils/Image.cs
aimerib: Good catch

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmSaveImageWS.py
mcmonkey4eva: the code for the 8bit section is different than it was before the PR. The red lines don't match the green lines, even though the only expected change is those lines get tabbed forward into the else block. Why is there a difference?

aimerib: Ah, I see.

Before it was:
```
        for image in images:
            i = 255.0 * image.cpu().numpy()
            img = Image.fromarray(np.clip(i, 0, 255).astype(np.uint8))
            pbar.update_absolute(step, SPECIAL_ID, ("PNG", img, None))
            step += 1
```
I noticed that this was grabbing the output of the tensor as a numpy array `image.cpu().numpy()` and scaling it to the 8bit color range. I was going to have to do the same operation for scaling it to 16bits, so it made sense in my mind to extract the numpy array into it's own variable, so I could just repeat the scaling with the appropriate ranges for 16bits.
I suppose another way I could have done the same thing is:

```
if bit_depth == "16bit":
    i = 65535.0 * image.cpu().numpy()
    img = self.convert_opencv_to_pil(np.clip(i, 0, 65535).astype(np.uint16))
else:
    i = 255.0 * image.cpu().numpy()
    img = Image.fromarray(np.clip(i, 0, 255).astype(np.uint8))
```
and this would have kept the syntax for 8bits same as before, and probably cleaner to review too. Is this what you're referring to? If so, I didn't really see this option until now, to be honest.

mcmonkey4eva: yes

aimerib: thank you for your patience. I'll fix this and push it up

mcmonkey4eva:
... This wasn't tested, was it? It doesn't work.

mcmonkey4eva:
So two problems:
- When pillow loads 48-bit images, in converts to 32bit (8bit RGBA). So you have to skip pillow processing. My above commit changes that.
- When ImageSharp loads the image, nope it fails. So... that'll need special handling too

mcmonkey4eva:
oh the ImageSharp error was my fault - after fixing that, the next issue was `== 16` (or `== 8`) in that code was wrong, as the actual value was `48`... I adapted it to `> 32` to allow both `48` and a hypothetical `64` to both be used. It seems to now work end-to-end

Issue #303, opened 2024-10-05
Separate inpaint prompt from general prompt in UI and metadata


Michoko92:
### Feature Idea

Hi,

Thank you for this great UI. I'd have a suggestion that would definitely improve inpaint refining process. Generally, we generate an image with a global prompt, then we use the EDIT feature to inpaint and fix some details. Then we send back the edited image to init image, to run the initial prompt again on it, and get a generally refined and coherent picture. Works wonders with Flux.

However, in the process of editing the image, we currently lose the initial prompt (as we use simpler prompts for inpainting, like "detailed hands", "detailed feet", and so on). Also in the generated images, the original prompt is lost in the metadata and is replaced by shorter inpainting prompts. I think it would be great that the inpaint prompt would be separate from the regular prompt, both in the UI and the metadata. It means that when we exit the EDIT UI, the prompt would be set back to the original full prompt, instead of keeping the inpainting prompt, that is generally useless outside of the editing process. That would make the back-and-forth process between generating images and inpainting them MUCH handier, and metadata preservation would be better (as each image would have both the original full prompt, and the prompt of the specific inpainting work that has been done on this generation).

Hope I made sense. 😉 Thank you!

aimerib:
If you're savvy with git, you can test the changes here: https://github.com/mcmonkeyprojects/SwarmUI/pull/307 
Any help testing changes to validate that they work and don't break anything else would be helpful.

Michoko92:
Hi, and thank you so much for having a look at it.

I'm not super good at github, but I tried to checkout your PR locally with 👍 

git fetch upstream pull/307/head && git checkout FETCH_HEAD

The .cs files seem to be updated, and the SwarmUI dll was recompiled. However, when I go to the Edit UI, it doesn't seem to have changed anything : the edit prompt and the regular prompt are still the same. Not sure what I'm doing wrong...

aimerib:
Check the Init Image params. If you expand that section, you should have an edit prompt there. You can leave your main prompts untouched, as if you add anything to the edit prompt, it will use that instead.

Michoko92:
[...]

PR #307, opened 2024-10-05
Add Init Image/Edit Image specific prompt and metadata for edit prompt

aimerib:
Adds a prompt box for init image params, specifically for using with edit image, but would work with init image as well. When a value is present in this prompt, the main prompt is ignored.

This will also add `editprompt` to the list of metadata.

This change is useful for when you have a long involved prompt as your main positive prompt, and you want to edit an image, but doesn't want to lose your prompt. This gives users a more ergonomic experience than both having a separate image2image tab AND having to replace your main prompt for every edit.

I suppose that this is sort of redundant if you are only doing img2img starting from a random picture dragged to the UI, but it is really useful when editing a generated image.

This was a pretty simple change, mostly to address https://github.com/mcmonkeyprojects/SwarmUI/issues/303 and my own needs. I'm more than open to discussing improvements or other approaches to this problem. I tested locally, and it seemed to work really well.

### DIFF
--- a/src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
+++ b/src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
@@ -391,7 +391,7 @@ public class WorkflowGeneratorSteps
                         });
                         g.FinalNegativePrompt = [zeroed, 0];
                     }
-                    if (!g.UserInput.TryGet(T2IParamTypes.Model, out T2IModel model) || model.ModelClass is null ||
+                    if (!g.UserInput.TryGet(T2IParamTypes.Model, out T2IModel model) || model.ModelClass is null || 
                         (model.ModelClass.CompatClass != "stable-diffusion-xl-v1"/* && model.ModelClass.CompatClass != "stable-diffusion-v3-medium"*/))
                     {
                         throw new SwarmUserErrorException($"Model type must be SDXL for ReVision (currently is {model?.ModelClass?.Name ?? "Unknown"}). Set ReVision Strength to 0 if you just want IP-Adapter.");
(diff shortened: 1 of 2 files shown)


src/Text2Image/T2IParamTypes.cs
```diff
@@ -385,6 +385,9 @@ public static void RegisterDefaults()
         InitImage = Register<Image>(new("Init Image", "Init-image, to edit an image using diffusion.\nThis process is sometimes called 'img2img' or 'Image To Image'.",
             null, OrderPriority: -5, Group: GroupInitImage, ChangeWeight: 2
             ));
+        EditPrompt = Register<string>(new("Edit Prompt", "Prompt to use specifically for editing images.\nThis will override the main prompt when generating edited images.",
+            "", OrderPriority: -4.9, Group: GroupInitImage, ViewType: ParamViewType.PROMPT
```
mcmonkey4eva, line 389: this needs either a toggleable or an ignoreif empty. It should probably also be advanced (the average person doesn't need this).

src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
```diff
@@ -333,7 +333,14 @@ public static void Register()
         #region Positive Prompt
         AddStep(g =>
         {
-            g.FinalPrompt = g.CreateConditioning(g.UserInput.Get(T2IParamTypes.Prompt), g.FinalClip, g.UserInput.Get(T2IParamTypes.Model), true, "6");
+            if (g.UserInput.TryGet(T2IParamTypes.EditPrompt, out string editPrompt) && !string.IsNullOrWhiteSpace(editPrompt))
```
mcmonkey4eva, line 336: this if block shouldn't be duplicating the entire code inside, it's only changing one variable

mcmonkey4eva:
It feels like it might be better here to just track the original prompt in metadata cleanly, rather than having a new parameter for edit prompt? When the init image was a generated image, copy the source prompt into metadata somewhere.
Not really sure.

Michoko92:
Just wanted to say that I've been using this PR for a few days now, and it is really very handy to make quick edits without losing the initial prompt. I hope it will eventually be merged, as it is a great addition.

Michoko92:
Is there still hope this PR will be merged or adapted, please? This is a very interesting feature, to be able to do inpainting iterations while keeping the original prompt of the main generation. Thank you! 🙂

mcmonkey4eva:
Closing for now as this has not been updated since 2024. Re-open if/when you want to continue.

Issue #122, opened 2024-08-03
"Upscale 2x" button that actually uses upscale?


CaptainVarghoss:
### Feature Idea

Some way to modify what process the 'upscale 2x' button does would be ideal, but simply having it use the upscale settings instead of whatever it's currently doing with the init image creativity.

Right off the bat, the label is misleading at best, being labeled as 'upscale 2x' one would expect it to use the upscale settings.
Second, the current method of upscaling does not use the upscaler model.
Third, as far as I can tell, I can't even modify the settings it uses, so the 0.4 'creativity' is destroying my images, even if the upscale was otherwise viable (it's not IMO)
Fourth, part of it's process is to completely regen the original image and then upscaling, wasting cycle time compared to simply turning on the upscale tab and reusing the seed.

In my mind, solving this could be done in a number of ways. 
At the minimum, a second button that simply uses the current upscaler settings (and maybe make the current one use the currently user set settings not whatever it uses)
At best, some way of making custom "turn these things on, generate with current settings, turn them back off" buttons would be amazing.

CaptainVarghoss:
Submitted a pull request for this here - #308 

I added a setting for the current "Upscale 2x" button to change how much creativity is being used (0.4 is heavy handed IMO). Ultimately left the function and default value alone for backwards compatibility and those that like the way it currently works.

Added a second button called "Refine Image" that reuses the current images seed and enables the "Refine / Upscale" section and generates an image using the existing settings, then disables the section and resets the seed to random (-1).

CaptainVarghoss:
Solved with merge of #312

PR #308, opened 2024-10-06
Upscale 2x setting and new button for Refine Upscale

CaptainVarghoss:
Currently the Upscale 2x button is hardcoded for a 0.4 InitImageCreativity. This adds a setting to the User Settings to change the value the button uses.

Added a new button "Refine Image" that generates the existing image with the current "Refine / Upscale" settings using the same seed.
Essentially the same as clicking the "Reuse Seed" button and enabling the "Refine / Upscale" toggle, generating the image, then turning the refine toggle back off and resetting the seed to random.

It seems the recommended way to upscale is by using the "Refiner / Upscale" section, so it made sense for it to have it's own button.

Original feature request (submitted by myself)
#122

### DIFF
--- a/src/Core/Settings.cs
+++ b/src/Core/Settings.cs
@@ -317,6 +317,9 @@ public class Settings : AutoConfiguration
             + "\nThe default is blank, which currently implies 'Use As Init,Edit Image,Star,Reuse Parameters'")]
         public string ButtonsUnderMainImages = "";
 
+        [ConfigComment("The ammount of image creativity used by the Upscale 2x button.")]
+        public double UpscaleCreativity = 0.4;
+
         [ConfigComment("If enabled, batch size will be reset to 1 when parameters are loaded.\nThis can prevent accidents that might thrash your GPU or cause compatibility issues, especially for example when importing a comfy workflow.\nYou can still set the batch size at will in the GUI.")]
         public bool ResetBatchSizeToOne = false;
 
--- a/src/Core/Settings.cs
+++ b/src/Core/Settings.cs
@@ -313,11 +313,11 @@ public class Settings : AutoConfiguration
         public bool AutoSwapImagesIncludesFullView = false;
 
         [ConfigComment("A list of what buttons to include directly under images in the main prompt area of the Generate tab.\nOther buttons will be moved into the 'More' dropdown.\nThis should be a comma separated list."
-            + "\nThe following options are available: \"Use As Init\", \"Edit Image\", \"Upscale 2x\", \"Star\", \"Reuse Parameters\", \"Open In Folder\", \"Delete\", \"View In History\""
+            + "\nThe following options are available: \"Use As Init\", \"Edit Image\", \"Upscale 2x\", \"Star\", \"Reuse Parameters\", \"Open In Folder\", \"Delete\", \"View In History\", \"Refine Image\""
             + "\nThe default is blank, which currently implies 'Use As Init,Edit Image,Star,Reuse Parameters'")]
         public string ButtonsUnderMainImages = "";
 
-        [ConfigComment("The ammount of image creativity used by the Upscale 2x button.")]
+        [ConfigComment("The amount of image creativity used by the Upscale 2x button.")]
         public double UpscaleCreativity = 0.4;
 
         [ConfigComment("If enabled, batch size will be reset to 1 when parameters are loaded.\nThis can prevent accidents that might thrash your GPU or cause compatibility issues, especially for example when importing a comfy workflow.\nYou can still set the batch size at will in the GUI.")]
--- a/src/wwwroot/js/genpage/main.js
+++ b/src/wwwroot/js/genpage/main.js
@@ -700,9 +700,25 @@ function setCurrentImage(src, metadata = '', batchId = '', previewGrow = false, 
                 'width': width * 2,
                 'height': height * 2
             };
-            mainGenHandler.doGenerate(input_overrides, { 'initimagecreativity': 0.4 });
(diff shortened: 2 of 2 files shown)


src/Core/Settings.cs
```diff
@@ -313,10 +313,13 @@ public class ThemesImpl : SettingsOptionsAttribute.AbstractImpl
         public bool AutoSwapImagesIncludesFullView = false;
 
         [ConfigComment("A list of what buttons to include directly under images in the main prompt area of the Generate tab.\nOther buttons will be moved into the 'More' dropdown.\nThis should be a comma separated list."
-            + "\nThe following options are available: \"Use As Init\", \"Edit Image\", \"Upscale 2x\", \"Star\", \"Reuse Parameters\", \"Open In Folder\", \"Delete\", \"View In History\""
+            + "\nThe following options are available: \"Use As Init\", \"Edit Image\", \"Upscale 2x\", \"Star\", \"Reuse Parameters\", \"Open In Folder\", \"Delete\", \"View In History\", \"Refine Image\""
             + "\nThe default is blank, which currently implies 'Use As Init,Edit Image,Star,Reuse Parameters'")]
         public string ButtonsUnderMainImages = "";
 
+        [ConfigComment("The amount of image creativity used by the Upscale 2x button.")]
+        public double UpscaleCreativity = 0.4;
```
mcmonkey4eva, line 321: This doesn't particularly make sense to have - Init Image Creativity is the value that gets used, the one in the JS is just a default that's lower than the standard default

src/wwwroot/js/genpage/main.js
```diff
@@ -700,9 +700,25 @@ function setCurrentImage(src, metadata = '', batchId = '', previewGrow = false,
...
     }, '', 'Runs an instant generation with this image as the input and scale doubled');
+    includeButton('Refine Image', () => {
+        toDataURL(img.src, (url => {
+            let input_overrides = {
+                'images': 1,
+            };
+            let toggler = getRequiredElementById('input_group_content_refineupscale_toggle');
+            toggler.checked = true;
+            triggerChangeFor(toggler);
+            reuseLastParamVal('input_seed');
```
mcmonkey4eva, line 715: This reuse/setrandom is not great here, use input overrides

mcmonkey4eva: ... also this code in general doesn't do what you'd expect a refine button to do. Nowhere here are you inputting the relevant image, for example.

CaptainVarghoss: > ... also this code in general doesn't do what you'd expect a refine button to do. Nowhere here are you inputting the relevant image, for example.

I'm not sure what you mean, this does exactly what I expect it to do with the current image.

mcmonkey4eva: Generate a picture of a cat, change the prompt box to say "waffle fries", then click the Refine button. What happens? Does it refine the image, or does it do something else entirely?

mcmonkey4eva: Alternately, drag a random image that didn't come from an AI at all into Swarm's interface, and click Refine. What does it do? Does it refine that image, or something else entirely?

CaptainVarghoss: Ok, I see what you mean. It doesn't work as intended when it's not used as intended. I'll have to fix that then

mcmonkey4eva: lol, if the intent of a button *on an image* that says "Refine" isn't to refine the image, you've done something very wrong somewhere in planning

CaptainVarghoss: No, that's exactly the intention, and that's what it does. Changing the prompt or the image itself before clicking the refine button is not the intended use, although it still does what it says it does. What I had not considered in my planning was that it would be used in that way.

The whole point is to use the current refine settings on the current image. It does that perfectly. What it does not do, is account for discrepancies between the current image and other settings since I didn't plan for someone to change anything before pressing the button.

mcmonkey4eva: You wrote code that adds a button every time any image is displayed in the center area of Swarm's interface, but only does anything of use if the image is fresh out of the generate button oven. The placement and your stated intention do not align. Technically you could get closer to your stated intention by moving the button somewhere else, but it's better to just allow refining images that aren't the most recently generated image.

CaptainVarghoss: My use case is exactly as you say.

Generate an image at a low resolution, like the image and want to upscale the image, click the 'Refine Image' button and get an upscaled version using the current refine settings.

It's intended to work exactly as the 'Upscale 2x' button, but using different parameters since the 'Upscale 2x' button does not produce the results I want, but I like the easy convenience of a single click button.

I'll admit my hacky fix might not be the best way to accomplish it, but it does exactly as it was intended, in the use case it was intended for.

I'm not sure where else I would put the button since in my use case I have the 'Refine Image' button shown by default in the center area, and the 'Upscale 2x' button falls under the 'more' list because I don't use it. 

Unfortunately I'm not familiar enough with the code at the moment to make it use the actually displayed current image and it's existing parameters and adding the refine parameters to that. However, if that's the solution then that's what I'll work on and figure out.

CaptainVarghoss:
For now, I'll close this request as it has unintended consequences under varying circumstances. I'll submit a different PR when I figure out a better way to handle it.

src/wwwroot/js/genpage/main.js
mcmonkey4eva: basically just feed the image as an init image, but override creativity to 0, and it oughtta work exactly how you're intending

Issue #175, opened 2024-08-16
[Feature Request] Auto refresh for Image History tab


kalebbroo:
### Feature Idea

There have been many times when I would love to have the image tab auto refreshing images. This way you can see them all in real time. I also share my instance with friends and like to watch together what we all create. 

I do not however need it to do this all the time so maybe a checkbox in the tab or user settings would be best. Or maybe it only does it if you have the tab open?

misakitchi:
Auto Refresh every XX minutes

PR #319, opened 2024-10-11
Image history browser auto refresh

CaptainVarghoss:
This covers feature request here #175 to add auto refreshing of the image history browser.

Added a setting to enable or disable (disabled by default)

Not sure what the resource hit would be on this if you have a massive image library or when generating large batches, so I added a warning on the setting. 

Not really sure if the generatehandler is the best location for this, but I couldn't find a logical place to stick it without some kind of "do this after generation" function similar to beforeGenRun()

### DIFF
--- a/src/Core/Settings.cs
+++ b/src/Core/Settings.cs
@@ -366,6 +366,9 @@ public class Settings : AutoConfiguration
         [ConfigComment("If true, the Image History view will cache small preview thumbnails of images.\nThis should make things run faster. You can turn it off if you don't want that.")]
         public bool ImageHistoryUsePreviews = true;
 
+        [ConfigComment("If true, will refresh Image History Browser after every new image generation.\nWARNING: This may cause heavy resource usage depending on your system and how many images exist.")]
+        public bool ImageHistoryAutoRefresh = false;
+
         [ConfigComment("Delay, in seconds, betweeen Generate Forever updates.\nIf the delay hits and a generation is still waiting, it will be skipped.\nDefault is 0.1 seconds.")]
         public double GenerateForeverDelay = 0.1;
 
--- a/src/wwwroot/js/genpage/generatehandler.js
+++ b/src/wwwroot/js/genpage/generatehandler.js
@@ -142,6 +142,9 @@ class GenerateHandler {
                         discardable[data.batch_index] = images[data.batch_index];
                         delete images[data.batch_index];
                     }
+                    if (getUserSetting('ImageHistoryAutoRefresh')) {
+                        imageHistoryBrowser.update();
+                    }
                 }
                 if (data.gen_progress) {
                     let thisBatchId = `${batch_id}_${data.gen_progress.batch_index}`;


mcmonkey4eva:
Yeah that's gonna be a huge redundant perf hit for no reason - this needs a much smarter impl than just raw refresh calls all the time

CaptainVarghoss:
> Yeah that's gonna be a huge redundant perf hit for no reason - this needs a much smarter impl than just raw refresh calls all the time

As far as I can tell, this is the same function that is called when a user clicks the 'refresh' button, so the only way I can see to make it any more efficient would be rewriting the way the image browser functions all together and implement some sort of additional caching.

Based on my own testing having it refresh when an image is generated doesn't cause any noticeable hit to resources (less than 0.15% cpu), versus clicking the refresh button after 10 or more images which causes a perceptible spike in cpu usage as it generates multiple previews at once.

I'm not sure the real world performance hit is 'huge', this only runs when a new image or batch finishes, it's not running all the time. It can also easily be disabled (default setting) in the settings if the user ends up having an actual performance concern with it.

The only mitigation I can think of is to make it only refresh if the image history browser is the current tab and also not hidden. This would also require additional code to update the history browser when it's un-hidden or the tab becomes current for it to appear 'automatic' to the user. It seemed a bit excessive to me as it could just be turned off if it actually hurts ones performance to have it refresh in the background, but if this is a solution that would be acceptable to merge I'll work on that.

brendanhoar:
It does depend upon what kind of connection one has to the backend. e.g. i believe there is an intent to have SwarmUI work well over mobile. Currently the refresh button pulls *all* of the metadata and thumbnails for everything indexed every time refresh is clicked.

As per @mcmonkey4eva, a future state would involve only sending changes on refresh calls. Or, alternately, maintaining updates locally for the session/account. 

This stuff overlaps with the other future intent of allowing separated account access/spaces as well, so simple solutions may not be  the best way to go at the moment.

mcmonkey4eva:
`As far as I can tell, this is the same function that is called when a user clicks the 'refresh' button,`
Yes, it is. It is a full refresh of the entire list of potentially thousand of images & their metadata. To add, _one image, or a batch of images_. The scale of how much is being processed relative to how much change is actually needed is pretty high.
Swarm is designed such that even relatively inefficient actions don't hurt that bad, but still as much as possible I'd prefer to seek more efficient and clean solutions to things.

`so the only way I can see to make it any more efficient would be rewriting the way the image browser functions all together and implement some sort of additional caching.`
There actually is or was already code to handle this feature in the past, it was just disabled cause it was more annoying than helpful. You can just add new entries to the browser file list when they appear. This is the original reason the `appendImage` function was made - it appended images directly to either the image history view or the batch view, but was swapped to just do batch later.
It'd need to be rebuilt to work with history browser now, as the history browser behaves differently than it used to.

An ideal solution here would probably be to integrate with the Browser code class, give it an `addIfVisible(entry)` that will add the entry, if it should be visible in the current view, to the correct spot within the current view, and automatically process metadata n wotnot. (As specifically opposed to the old `appendImage` trick just directly shoving new html into a div)

CaptainVarghoss:
Thank you for the explanation. This might be a bit beyond what I can do with my limited knowledge of JS at this point, but I'll poke around a bit more and see if I can figure it out.

PR #413, opened 2024-11-13
Minor fix for host network line in docker compose

PsychoLogicAu:


mcmonkey4eva:
Could you please explain this change and/or link relevant docs?

PsychoLogicAu:
The syntax as was present is invalid and rejected by docker compose:
```
validating /path/to/docker-compose.yml: services.swarmui Additional property network is not allowed
```

Docs for `network_mode`: https://docs.docker.com/reference/compose-file/services/#network_mode

In short, `network_mode: host` 'Gives the container raw access to the host's network interface.' which appears to have been intended here.

Docs for `networks`: https://docs.docker.com/reference/compose-file/services/#networks
is for defining which networks a service container is attached to (these in turn are defined under the top-level element `networks` in the docker compose file).

I imagine an equivalent version using `networks:` would be:
```
services:
  swarmui:
    #...
    networks:
      - host_network

networks:
  host_network:
    driver: bridge
```


An equivalent version using networks:

mcmonkey4eva:
Ah, nice. Thank you!

PR #434, opened 2024-11-21
Fix for gridgen fullscreen preview failure

brendanhoar:
It appears the object 'data' does not exist. Perhaps 'metadata' was what was needed?

In any case, full screen previews were broken, this seems to fix it.

The code is a little more testy test test, but I can reduce it back to just changing the object name to metadata if you want.

B


mcmonkey4eva:
Heck, yep, bad edit on the name there.

Your changes were wrong but yeah the name needed to be swapped, it was a var rename not properly propagated

Issue #439, opened 2024-11-25
'Mask shrink grow' doesn't behave as expected when inpainting with 'Flux fill' model


willhsmit:
### Expected Behavior

When using a Flux fill model with an inpainting mask with 'Mask shrink grow' turned on and set to a value, Swarm uses a cropped portion of the image near the mask as input to the generation, generates a new image of the cropped portion with the masked area repainted, and recomposites onto the fullsize original image.
 

### Actual Behavior

Swarm uses the fullsize image as input to the generation, generates a new version of the full image with the masked area repainted, resizes the fullsize image down to the cropped area around the mask, and composites it in. This results in portions of the fullsize image being repeated inside the mask, and distorts the aspect ratio of the mask.

Not sure I've explained it well, but example:
Input image:
![389207266-6da62741-36a1-4fd6-b523-4cbeb9c9bb86-crop](https://github.com/user-attachments/assets/bb2266a2-2acb-4a91-a155-a773c00a42c7)
Mask:
![mask](https://github.com/user-attachments/assets/e0b6b5e1-f2d1-46fe-9e96-7b69ca144aa8)

Prompt: Man with jacket
seed: 2003358640
model: flux1-fill-dev-Q4_K_S.gguf

Output:
![1636-a man wearing a jacket-flux1-fill-dev-Q4_K_S-2003358640](https://github.com/user-attachments/assets/1861c302-160e-4667-86ce-d225d4db3b95)

What you can see is that it's repeated the original 992x992 image inside the mask. 

### Steps to Reproduce

1. Pick a flux fill model
2. Drag an image into the UI 'Use as Init', 'Edit image'
3. Draw a mask on a portion of the image
4. Set 'Init creativity to 1', 'Mask Shrink Grow' to on and 32.
5. Generate

During generation, you can see the full original image in the preview, whereas with other models during the preview phase when 'Mask shrink grow' is on, you only see a cropped area around your mask. At the end you get the effect shown in the screenshots.

### Debug Logs

== SwarmUI logs 2024-11-24 17:00 ==
17:00:51.981 [Info] User local requested 1 image with model 'flux1-fill-dev-Q4_K_S.gguf'...
17:02:20.067 [Info] Generated an image in 2.16 (prep) and 83.84 (gen) seconds

mcmonkey4eva:
fixed

willhsmit:
For posterity, this is the current state of the above example, after the two commits:

[...]

PR #443, opened 2024-11-28
Scale the mask to match the scaled image when using mask shrink with flux fill

willhsmit:
To fully fix https://github.com/mcmonkeyprojects/SwarmUI/issues/439, I think the mask also needs to be scaled before passing into the conditioning function.

I wasn't able to find an existing scaled copy of the mask, or an equivalent to SwarmImageScaleForMP that works on masks, so there's some repeated code here. There may be a better way of reusing SwarmImageScaleForMP or work that's already been done scaling the image.

### DIFF
--- a/src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmMasks.py
+++ b/src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmMasks.py
@@ -268,7 +268,7 @@ class SwarmMaskScaleForMP:
     CATEGORY = "SwarmUI/masks"
     RETURN_TYPES = ("MASK",)
     FUNCTION = "scale"
-    DESCRIPTION = "Scales an mask to a target width and height, while keeping the aspect ratio."
+    DESCRIPTION = "Scales a mask to a target width and height, while keeping the aspect ratio."
 
     def scale(self, mask, width, height, can_shrink):
         mpTarget = width * height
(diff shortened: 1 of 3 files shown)


willhsmit:
With this, I get fairly good results for the example in https://github.com/mcmonkeyprojects/SwarmUI/issues/439:

1807-Man with jacket-flux1-fill-dev-Q4_K_S-2003358640

mcmonkey4eva:
Valid identification of where the problem was, wrong solution to it though -- it just needed to be the CroppedMask instead of the full mask.

willhsmit:
Huh. I swear I tried doing that first. 

Results are exactly as I expected now. Thanks!

FurkanGozukara:
awesome so i am assuming something fixed :D 

do we have example?

willhsmit:
This was for https://github.com/mcmonkeyprojects/SwarmUI/issues/439 but was rejected in favor of a better fix https://github.com/mcmonkeyprojects/SwarmUI/commit/5e303de2f58b36ae1d5d775caf60dfe0a3687677. 

439 has reproduction steps for an example of what was going wrong, I commented on 439 with what the reproduction steps give me after the latest fix.

Mask Shrink Grow's a feature that you've probably used a long time on other models anyway, but a more realistic demo of how to use 'Mask shrink grow' with Flux fill would probably be:
(1) generate a 1024x1024 photorealistic or anime picture with a character using non-fill flux schnell/dev
(2) you are generally happy with your gen, but some detail like a hand is a mess
(3) you mask the hand to fix it with an inpainting model like Flux fill. Maybe the masked area ends up being like 50 pixels square
(4) Generating with the whole image as a canvas for inpainting continues to struggle to get the fine detail right in the 50x50 hand area.
(5) Using 'Mask shrink grow', find a grow value that is large enough that the Flux fill can access surrounding context to place the hand correctly, but small enough that the hand takes up more of the image and Flux fill can focus more on the detail. For example, a grow value of 96 with a 50x50 mask means the UI is scaling a ~242x242 space up to 1024x1024, growing your 50x50 mask into a ~200x200 area that gives the details more focus, before shrinking and recompositing again.
(6) Generate and hopefully get a better hand.

Side comments:
(1) Particularly useful if the original image wasn't 1024px; if it's much smaller the model will struggle even harder to get small details right without using 'Mask shrink grow' to zoom in; if it's much bigger, it will take a lot of time and VRAM to use a full 3000+ image as the canvas just to infill a small part of it unless you use 'Mask shrink grow'.
(2) A hack alternative to tuning the 'Mask shrink grow' value carefully is to set it to a very low value, and then add little extra outlying pixel-sized dots to your mask to define exactly where you want the boundaries of the canvas rectangle to be.

FurkanGozukara:
@willhsmit so this is generating masked area in bigger resolution and then automatically shrinking back into that small area basically copy pasting right?

excellent

PR #459, opened 2024-12-03
Import rembg locally to prevent slow imports

d8ahazard:
Sometimes, for some reason, importing rembg can add like 40s to the startup time for ComfyUI. Let's put that load in the actual function, at least until I can fork and fix rembg.

This is more or less identical to the fix I suggested for comfy_mtb:

https://github.com/melMass/comfy_mtb/pull/216

### DIFF
--- a/src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyExtra/SwarmRemBg.py
+++ b/src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyExtra/SwarmRemBg.py
@@ -1,7 +1,6 @@
 from PIL import Image
 import numpy as np
 import torch
-from rembg import remove
 
 class SwarmRemBg:
     @classmethod
@@ -17,6 +16,8 @@ class SwarmRemBg:
     FUNCTION = "rem"
 
     def rem(self, images):
+        from rembg import remove
+
         output = []
         masks = []
         for image in images:


mcmonkey4eva:
why must `rembg` continually be the most cursed lib of all the python libs, when it's nowhere near the most unusual in what it does

d8ahazard:
> why must `rembg` continually be the most cursed lib of all the python libs, when it's nowhere near the most unusual in what it does

LOL, you have NO idea how long this has been bugging me. Originally, it was the comfy_mtb custom node causing a 42s+ delay on first ComfyUI startup. Fix that, it moves to the node for SwarmUI. :P

Hopefully, I don't have any other custom nodes that use rembg. :D 

Thanks for the quick merge, and nice talking to you again. ;)

FurkanGozukara:
I always see rembg in logs but how do we exactly use it in SwarmUI @mcmonkey4eva 

Also nice to see you here @d8ahazard

southbayjay:
> I always see rembg in logs but how do we exactly use it in SwarmUI @mcmonkey4eva
> 
> Also nice to see you here @d8ahazard

I just did a video on it. Very easy to use.. 

https://youtu.be/PoAFPrpG4oA?si=az3HaeDI5O74rnix

FurkanGozukara:
@southbayjay great one. can we use it on existing images to remove background?

southbayjay:
> @southbayjay great one. can we use it on existing images to remove background?

Thanks! Yes but only if you have created that image with Swarm, just load the image to the workspace and click "reuse params" to generate it with the remove background enabled.

FurkanGozukara:
> > @southbayjay great one. can we use it on existing images to remove background?
> 
> Thanks! Yes but only if you have created that image with Swarm, just load the image to the workspace and click "reuse params" to generate it with the remove background enabled.

so that is actually no :D ye i saw how you make reuse params. thanks for replies

mcmonkey4eva:
you can set any image as the init image, creativity to 0, and enable remove background

southbayjay:
ohhh fantastic!!

FurkanGozukara:
> you can set any image as the init image, creativity to 0, and enable remove background

wow nice way

PR #477, opened 2024-12-18
Update it.json

medoro:
Make some corrections: VAE was auto-trasnlated as United Arab Emirates O_o

Changed a few other words, I'll check other translation later..


mcmonkey4eva:
`VAE was auto-trasnlated as United Arab Emirates`

I love AI.

(Also I need to do a general regen of the translation files, haven't been updated in months...)

Issue #494, opened 2024-12-28
Add option to add unique index to output file name in ImageBatch 


martinken:
### Feature Idea

Use case, running image batch on a set of input images to extract/crop faces and save them as images. One input image can produce multiple output images, one per face.

Currently the ImageBatch extension will fire multiple times if needed for an input image, but writes to the output directory with the same filename (based on the input image) ala https://github.com/mcmonkeyprojects/SwarmUI/blob/bae492dc2c7b7ac88ee340252bbbb9114c75bb5e/src/BuiltinExtensions/ImageBatchTool/ImageBatchToolExtension.cs#L199 

Feature would be to add an option to append an index to that output filename (possibly the one already in the metadata object)

mcmonkey4eva:
added

martinken:
tested locally and it works for my use case, thanks!

PR #498, opened 2024-12-30
FEAT: support more than one output image per input image

martinken:
For the image batch extension, this change makes it so that when a workflow produces more than one output image for an input image, the output images after the first will be appended with a _1 _2 etc.  For example:

input -> output
File1.png -> File1.png, File1_1.png, File1_2.png   // for three generated images
File2.png -> File2.png   // only one output
File3.png -> File3.png, File3_1.png  // two generated

Addresses Issue https://github.com/mcmonkeyprojects/SwarmUI/issues/494

### DIFF
--- a/src/BuiltinExtensions/ImageBatchTool/ImageBatchToolExtension.cs
+++ b/src/BuiltinExtensions/ImageBatchTool/ImageBatchToolExtension.cs
@@ -115,10 +115,6 @@ public class ImageBatchToolExtension : Extension
         {
             string fname = file.Replace('\\', '/').AfterLast('/');
             int imageIndex = batchId++;
-            // the following two variables are used to handle when one input file
-            // produces multiple output images.In that case the first image is output
-            // as normal "output_dir/preExt.ext" but any images after that will be of the
-            // form "output_dir/preExt_lastFileCount.ext" 
             string lastFilePreExt = "";
             int outputCountForInput = 0;
             removeDoneTasks();
@@ -203,7 +199,6 @@ public class ImageBatchToolExtension : Extension
                     ext = properExt;
                 }
                 string outputFilePrefix = $"{preExt}";
-                // if this is the first output image for an input image initialize
                 if (outputFilePrefix != lastFilePreExt)
                 {
                   lastFilePreExt = $"{preExt}";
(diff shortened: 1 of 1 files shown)


src/BuiltinExtensions/ImageBatchTool/ImageBatchToolExtension.cs
```diff
@@ -18,6 +18,18 @@ public class ImageBatchToolExtension : Extension
 {
     public static PermInfo PermUseImageBatchTool = Permissions.Register(new("imagebatcher_use_image_batcher", "[Image Batch Tool] Use Image Batcher", "If true, the user may use the Image Batcher tool. Only makes sense for localhost users.", PermissionDefault.ADMINS, Permissions.GroupUser));
 
+    //   the following two statics are used to handle when one input file
+    //   produces multiple output images.In that case the first image is output
+    //   as normal "output_dir/preExt.ext" but any images after that will be of the
+    //   form "output_dir/preExt_lastFileCount.ext" 
+
+    // the preext of the last file written out
+    public static string lastFilePreExt;
```
mcmonkey4eva, line 27: this should not be static at all 0.o

src/BuiltinExtensions/ImageBatchTool/ImageBatchToolExtension.cs
```diff
@@ -60,6 +72,7 @@ public static async Task<JObject> ImageBatchRun(WebSocket socket, Session sessio
             await socket.SendJson(new JObject() { ["error"] = "Image batch needs to supply the images to at least one parameter." }, API.WebsocketTimeout);
             return null;
         }
+        lastFilePreExt = ""; // initialize to empty when starting a batch
```
mcmonkey4eva, line 75: all the excess `//` comments are not needed

src/BuiltinExtensions/ImageBatchTool/ImageBatchToolExtension.cs
martinken: Oops, sorry, didn't see the for loop scope it was inside of. Moved to that scope. Hopefully that is the right approach. Seems to work in my local tests. If you want the commits squashed etc let me know. Bit rusty.

src/BuiltinExtensions/ImageBatchTool/ImageBatchToolExtension.cs
```diff
@@ -115,6 +115,12 @@ void removeDoneTasks()
         {
             string fname = file.Replace('\\', '/').AfterLast('/');
             int imageIndex = batchId++;
+            // the following two variables are used to handle when one input file
```
mcmonkey4eva, line 118: these `//` comments are still really not needed

martinken: will yank

mcmonkey4eva:
The code here is still wrong (the if check is redundant)

But more importantly, I realize a bit late - this code section is potentially multithreading sensitive, and needs a solution written with an understanding of thread-safe coding. I was wrong to have put an _Easy PR_ label on that issue, my mistake there, I didn't realize that would apply.

I've gone ahead and implemented it correctly myself via the commit above

PR #511, opened 2025-01-06
Delete Selected Lora Button.

JJHACKER:
Added features:
1- Delete Selected Lora Button. (Delete all selected lora in one click)

Bug Fixed:
1- After removing lora manually with delete model , the lora list tab not getting updated.


mcmonkey4eva:
This is not the correct approach here imo - I want a proper Browser multi-select and bulk edit interface (`browers.js`), whereas this PR is:
(A) janking the LoRA selection system not a browser system. Being a browser system lets it apply to all browser lists (models, image history, etc) not just loras
(B) specifically delete? rather than generic (it should use a generic system to allow for more dynamic buttons to be easily added, notably for models you'd also want folder move and stuff like that)
(C) has the button in a rather odd location. I think the multi-edit controls should appear at the top of the browser view, maybe next to or below the line that has Format/Depth/Filter/etc.

I'm not sure if the LoRA selection should swap to using the generic browser selection, or if multi-select should be separate. I'm leaning towards separate - you have to explicitly click a "Multi-Select" button to start multi-selecting, and that doesn't enable/disable models, it just lets you select for bulk editing. The reasoning here being bulk multi-edits should not be easy to do _by accident_, which could happen if the normal model toggle counts as a multiselect.

PR #529, opened 2025-01-14
Generate forever with variation seed

aazsamir:
When generating forever, UI checks that the seed is not set to something other than -1, as it would be unnecessary to generate the same image over and over again.

However, the user can set the variation seed and it should keep generating images if the variation seed is set to random.

If variation filter bar switch isn't checked, variation would equal null, so it works as expected.


mcmonkey4eva:
how the heck did you add ltxtricks as a submodule 0.o

aazsamir:
sorry, I don't know how I missed it, blind `git add .`
I force pushed without it^^

Issue #597, opened 2025-02-17
Recomposite color correct improvements (broken out of Small Improvements)


willhsmit:
### Feature Idea

Opening a separate feature thread for the 'Recomposite color correct' for flux fill feature mentioned in #550. Original description:

> "recomposite color correct" to make flux fill work well -- line up pixels, select those with mask=0, ensure there's more than a minimum pixel count, and then measure what color correction is needed, take the average, and apply it to the full image -- see https://discord.com/channels/1243166023859961988/1243185862234210389/1327390619068530729 for my attempt at this. Unfortunately flux vae is dumdum inconsistent so even that's not good enough. probably need a full multiaxis gradient for the color shift. Might be a smarter way to do it? idk

The discord link has code that compares the old and new pixels outside the mask, computes the average difference between them in HSV, and applies that uniformly to the pixels inside the mask. The discord also suggests some areas of improvement:

> problem is, the color shift from the flux vae is inconsistent. I tried to do a global color correct, but if you fix the left, it'll just be wrong a different direction on the right. Maybe a, like, multi-dimension gradient from all edges would do nicely? ie do the shift estimate 4 or 8 times, for each different direction, and then smoothblend between them. I think the real solution is just throw up your hands and use a blurred mask, or don't recomposite at all :(

willhsmit:
First thing, I ported mcmonkey's original code from the discord into a test fork (https://github.com/willhsmit/SwarmUI) to test it out.

Here's an example of it working well on an image where the fill boundary is mostly same color - you can see an ugly seam around the 'sample' text in Before, that more or less goes away with correction.

Before:

![Image](https://github.com/user-attachments/assets/4238f2c7-1b7a-4f3e-b95f-6646d982431c)

After:

![Image](https://github.com/user-attachments/assets/335d05a3-f045-48e6-acee-02953c030012)


And an example of it working less well where the fill area has a mixture of white, purple, and black.

Before:

![Image](https://github.com/user-attachments/assets/5118af04-16ef-41cc-893d-bd1c374b978e)

After:

![Image](https://github.com/user-attachments/assets/146b4bab-0965-4979-bd02-2e18e49047d1)

willhsmit:
The original proposed solution to this was some version of using the gradient - modeling the diff not as a uniform value but as some kind of function over space to a diff, trying to blend in the values from different sides with a gradient.

I think this could get hairy for a couple reasons:
[...]

PR #598, opened 2025-02-18
Add code to do color compositing on inpainting

willhsmit:
This is intended as a fix for https://github.com/mcmonkeyprojects/SwarmUI/issues/597, which might take some further iteration.

Adds a UI parameter under 'Init Image' called 'Color Correction Behavior' with values 'None', 'Uniform', and 'Linear'.
* None is the default and is intended as a noop from current
* Uniform implements the behavior from https://discord.com/channels/1243166023859961988/1243185862234210389/1327390619068530729: find the mean difference between source and dest for each of HSV and apply it to source.
* Linear attempts to model the dest as a linear function of source and apply the function to source. Experimentally, I found that Hue is rarely improved and sometimes made a lot worse by correction, and decided to leave Hue uncorrected. Typically, the derived function will increase the saturation and contrast of the image somewhat.

You mentioned putting it behind a parameter; I assume you meant a UI parameter and not a server/user parameter that hides the UI parameter - I didn't see a lot of prior art for that but am open to fixing it.

I left the Uniform method as an option since I think the Linear option is not always better than it and people may want to experiment. I mostly left the code for Uniform as it was in the discord, but am open to:
* Trying the same removal of Hue as Linear
* Removing the .remainder() from the clamp at the end. I took this out of Linear because it sometimes produces black spots in the middle of a bright white. This happens occasionally but less often in Uniform; I wasn't sure if the remainder() was in there for an important reason.

### DIFF
--- a/src/Text2Image/T2IParamTypes.cs
+++ b/src/Text2Image/T2IParamTypes.cs
@@ -512,7 +512,7 @@ public class T2IParamTypes
             "false", IgnoreIf: "false", FeatureFlag: "controlnet", Permission: Permissions.ParamControlNet, VisibleNormally: false
             ));
         GroupVideo = new("Image To Video", Open: false, OrderPriority: 0, Toggles: true, Description: $"Generate videos with Stable Video Diffusion.\n<a target=\"_blank\" href=\"{Utilities.RepoDocsRoot}/Features/Video.md\">See more docs here.</a>");
-        static bool isVideoClass(string id) => id.Contains("stable-video-diffusion") || id.Contains("lightricks-ltx-video") || id.Contains("-video2world");
+        static bool isVideoClass(string id) => id.Contains("stable-video-diffusion") || id.Contains("lightricks-ltx-video") || id.Contains("-video2world") || id.Contains("-i2v");
         VideoModel = Register<T2IModel>(new("Video Model", "The model to use for video generation.\nThis should be an SVD (Stable Video Diffusion) model.\nNote that SVD favors a low CFG (~2.5).\nLTXV prefers a CFG around 3.",
             "", GetValues: s => CleanModelList(Program.MainSDModels.ListModelsFor(s).Where(m => m.ModelClass is not null && isVideoClass(m.ModelClass.ID)).Select(m => m.Name)),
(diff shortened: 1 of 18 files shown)


mcmonkey4eva:
The parameter should be... somewhere else. Honestly not sure where, maybe shoved under Advanced Sampling?

Because it's needed for `<segment:` usage especially, not just `init image`.
That or both an init image one, and a `Regional Prompting` one.

src/Text2Image/T2IParamTypes.cs
```diff
@@ -443,6 +443,9 @@ public static void RegisterDefaults()
         UseInpaintingEncode = Register<bool>(new("Use Inpainting Encode", "Uses VAE Encode logic specifically designed for certain inpainting models.\nNotably this includes the RunwayML Stable-Diffusion-v1 Inpainting model.\nThis covers the masked area with gray.",
             "false", IgnoreIf: "false", Group: GroupInitImage, OrderPriority: -3.2, IsAdvanced: true
             ));
+        ColorCorrectionBehavior = Register<string>(new("Color Correction Behavior", "Experimental: How to correct color when compositing a mask.\n'None' = Do not attempt color correction.\n'Uniform' = Compute a fixed offset HSV correction for all pixels.\n'Linear' = Compute a linear correction that depends on each pixel's S and V.",
+            "None", Toggleable: true, IsAdvanced: true, GetValues: (_) => ["None", "Uniform", "Linear"], OrderPriority: -3.1, Group: GroupInitImage
```
mcmonkey4eva, line 447: can do `IgnoreIf: "None"` instead of toggleable

willhsmit: Done.

willhsmit:
> The parameter should be... somewhere else. Honestly not sure where, maybe shoved under Advanced Sampling?
> 
> Because it's needed for `<segment:` usage especially, not just `init image`. That or both an init image one, and a `Regional Prompting` one.

I can't think of a better spot for the parameter than Advanced Sampling; I put it under some VAE-related parameters.

I tried doing some segment tests, the seam and the correction effect are pretty subtle. I'll add a couple to the bug thread. 

--

I ended up removing remainder() and also Hue from the 'Uniform' algorithm; there are some pretty clear examples from testing where those made things worse.

PR #630, opened 2025-03-04
Extend Video should use Video Settings

maedtb:
Extend Video was using Image CFG/Steps instead of Video CFG/Steps. Renamed `cfg` and `steps` local variables to `videoCfg` and `videoSteps` for added clarity.

Additionally: removed the default `videoCfg`, as passing `null` will use smarter defaults depending on the model. If this undesirable, i.e. a higher number like 7 was more desirable for _all_ models, I can revert this bit.


mcmonkey4eva:
This PR (A) does a few random changes (variable renames outside of affected code), and (B) doesn't do what you think it does -- `Video Steps` and `Video CFG` are `Image to Video` param group, ie it's not even a valid/present param if ImageToVideo isn't in use.
Arguably it would make sense to have a layered fallback - imagetovideo params if set, or general value if not?
The expectation is your input before extend is always a video, generated via _either_ t2v or i2v (or just tossed into the init image input)

PR #634, opened 2025-03-06
Replace `FreneticUtilities.FreneticExtensions.XFast` methods

GuyPago:
I've looked into these "Fast" methods, provided by `FreneticUtilities.FreneticExtensions`, for example `ToLowerFast` and I was fairly disappointed to see `Char[]` allocation, while dotnet's implementation uses `Span<char>`.

So I wanted to see if my intuition was wrong by running the following benchmark:
```C#
    [GlobalSetup]
    public void Setup()
    {
        // Generate a huge string (1 million characters)
        hugeString = new string(Enumerable.Repeat("AaBCdf DEshkFGe HIJKcdLMNO gePQRddSTUv SJD VWiyXYZ", 20000)
            .SelectMany(s => s).ToArray());
    }

    [Benchmark]
    public string ToLower() => hugeString.ToLower();
    [Benchmark]
    public string ToLowerFast() => hugeString.ToLowerFast();
    [Benchmark]
    public string[] Split() => hugeString.Split(' ');
    [Benchmark]
    public string[] SplitFast() => hugeString.SplitFast(' ');
```
Which yields the results:

  [Host]     : .NET 8.0.13 (8.0.1325.6609), X64 RyuJIT AVX2
  DefaultJob : .NET 8.0.13 (8.0.1325.6609), X64 RyuJIT AVX2

| Method      | Mean       | Error     | StdDev    | Median     | Gen0     | Gen1     | Gen2     | Allocated |
|------------ |-----------:|----------:|----------:|-----------:|---------:|---------:|---------:|----------:|
| ToLower     |   323.7 us |  10.73 us |  31.63 us |   312.3 us | 283.6914 | 283.6914 | 283.6914 |   1.91 MB |
| ToLowerFast | 1,390.3 us |  23.50 us |  20.83 us | 1,387.5 us | 597.6563 | 597.6563 | 597.6563 |   3.81 MB |
| Split       | 6,694.7 us | 120.43 us | 106.76 us | 6,698.2 us | 539.0625 | 531.2500 | 195.3125 |   4.88 MB |
| SplitFast   | 6,889.0 us | 116.52 us | 108.99 us | 6,901.8 us | 546.8750 | 539.0625 | 203.1250 |   4.88 MB |

As you can see, while `Split` didn't show much difference, `ToLowerFast` indeed allocates the huge string twice, while `ToLower`  allocates once as it loops through `Span<char>`, avoiding additional heap allocation.
`ToLower` is $\times 4.3$ faster and allocates half as much. This is fairly noticeable when working with big data scale tokens. 

One can open a followup PR that adds a custom analyzer to warn about `X.Fast` usage.

### DIFF
--- a/src/Accounts/User.cs
+++ b/src/Accounts/User.cs
@@ -92,7 +92,7 @@ public class User
     {
         lock (SessionHandlerSource.DBLock)
         {
-            return SessionHandlerSource.GenericData.FindById($"{UserID}///${dataname}///{name.ToLowerFast()}")?.Data;
+            return SessionHandlerSource.GenericData.FindById($"{UserID}///${dataname}///{name.ToLower()}")?.Data;
         }
     }
 
@@ -129,7 +129,7 @@ public class User
             {
                 return;
             }
-            SessionHandler.GenericDataStore dataStore = new() { ID = $"{UserID}///${dataname}///{name.ToLowerFast()}", Data = data };
+            SessionHandler.GenericDataStore dataStore = new() { ID = $"{UserID}///${dataname}///{name.ToLower()}", Data = data };
             SessionHandlerSource.GenericData.Upsert(dataStore.ID, dataStore);
         }
     }
@@ -139,7 +139,7 @@ public class User
     {
         lock (SessionHandlerSource.DBLock)
         {
-            return SessionHandlerSource.GenericData.Delete($"{UserID}///${dataname}///{name.ToLowerFast()}");
+            return SessionHandlerSource.GenericData.Delete($"{UserID}///${dataname}///{name.ToLower()}");
         }
     }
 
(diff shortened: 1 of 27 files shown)


mcmonkey4eva:
When that `Fast` code was written, a few dotnet versions again, `ToLower` was noticeably quite slow -- since you've brought this up, I'll do some testing, and if it's actually significantly faster now I'll see about getting `Fast` variants revised or replaced as relevant.

Aside from being faster, they are also written to be *more correct*. The native methods have several _localization errors_. That is to say: in certain localities, they return different results, and/or misbehave in general around non-English characters. This is a known core fault in the dotnet native methods that I'm quite sure they haven't fixed still.
FreneticUtilities has two corrections for this: one is a SpecialTool to disable most dotnet localization behaviors, and the other is these method variants that are explicitly implemented to behave in exactly one predictable manner.
The native methods also in some cases misbehave around "extremely special characters", such as `\0`, which is an issue so severe I'd call it a security risk for any code that relies on it - user strings with sufficiently bad data slipped in can break any string processing using some of the dotnet builtin methods.

GuyPago:
Thank you for elaborating

GuyPago:
Great overview! Also, I've seen your new commit for `Fast` and was pretty amused to see that you also adopted that `goto` statement idea :)

I got my masters in Data Science last year, I believe most of my projects are from there, but I've been working as a software dev for the past 3 years, half of it using .NET.

PR #637, opened 2025-03-08
Fix pip install -U in Installation.cs

darmach:
Fixes missing space before -U parameter for pip install causing pip install to fail claiming that install-U is wrong option :)


mcmonkey4eva:
Oh, oof, thank you.

*Can you tell the AMD route is dangerously undertested?*

If you encounter other issues on AMD cards please do submit more pull requests, it's very hard finding active contributors that use AMD.

darmach:
@mcmonkey4eva Sure thing, another thing out of the box I noticed, was wrong torch module version - installed was 2.4.0 while it expected 2.0.0. Not sure where it is declared yet.

PR #643, opened 2025-03-11
Colab slightly more straightforward, Spanish Readme and Modified Spanish UI Translation

usamireko:
Pretty much what the title says

The UI translation is not 100% done so I´ll update in the future, also the ReadMe but with less priority

image

The Colab makes it sligthly easier to play with it


mcmonkey4eva:
Please separately submit the readme and Spanish translation edits.

Don't resubmit the colab file change as is though, that's full of errors and I'm not sure why you changed it at all. I can't imagine what the goal of those sudo chmods is

PR #653, opened 2025-03-15
Add Podman Compatibility

little-file:
This pull request adds support for the Podman container manager to the project. The existing `launch-docker.sh` script has been adapted to work with Podman.

**Note:**

This change does not affect existing Docker users. It simply provides an additional launch option for Podman users.

### DIFF
--- a/launch-podman.sh
+++ b/launch-podman.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+podman build -t stableswarmui .
+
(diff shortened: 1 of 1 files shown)


mcmonkey4eva:
That's a docker script from the stability repo from a year ago, it does not match the current docker setup at all

mcmonkey4eva:
might be better to just add a podman section to the docker doc https://github.com/mcmonkeyprojects/SwarmUI/blob/master/docs/Docker.md ?

PR #659, opened 2025-03-18
Make w,h constants accurately reflect pre-defined aspect ratios (also check out my branch's wiki)

edtomb:
The resolutions for the predefined aspect ratio selections were scaled to the nearest multiples of 16. Not sure if this was intentional or for compatability reasons or something. The rounding error is not really noticeable at 512x512, but it gets worse at 1024x1024, and even worse if you go to upscale an image. I noticed this when I generated a 16:9 (actually 7:4) image and upscaled it to 1920x1080 and had black bars. I don't know enough about graphics processing to determine whether the performance overhead is worth the error, but I think it's worth at least having an option to pick an exact aspect ratio. 

Also, I made a wiki. I'm still working on it, and looking into ways to possibly keep it synced up with SwarmUI/docs/ with github actions or something. Anyway, let me know what you think. And thanks for the really great project


mcmonkey4eva:
Most models *can't* generate more precisely than x16 (latent subres is almost always 8x8, usually at least a 2x2 downscale in the backbone as well). Some can't go below x64 (eg unets have 8x8 downscale on top of the 8x8 latent scale yielding bottom blocks that are 64x64 pixels per latent line), others struggle with quality below x64 (even where capable of generating at more precise resolutions, they're often trained with images in x64 batched sets).
This is an issue with *latent diffusion model architecture*, not with SwarmUI's presentation of the stable resolution options.

The resolutions in the base set are based on the actual training sizes used for SDXL and SD3 (and many others), so they are exactly correct at 1024 scale models.

if you'd like to contribute to helping document SwarmUI, make PRs to the docs folder. However please take care to ensure you understand what you're writing documentation for beforehand. Maybe post in the Swarm Discord?

edtomb:
Understand. I was pretty sure there was an actual reason that it was coded that way, so thanks for explaining.

PR #660, opened 2025-03-18
Save sorting & view settings in cookie storage to survive restarts

AndrejMitrovic:
This preserves the following options in the default 90-day cookie storage:
- format option (Cards, Thumbnail, Lists, etc)
- sort by option
- reverse sort option
- depth

Tested it and it seems to work ok for me.

I see there's a few more options we could save but I'm not sure if they're worth storing. For example `filter`, but this is probably transitory.

----

There's a bit of key path copy-pasta here. Let me know if you want me to refactor this.

Cheers~


mcmonkey4eva:
localStorage already survives restarts. That's why that data is there.

AndrejMitrovic:
Oh, I think they were not surviving when using private browsing sessions. No worries, thanks~

PR #664, opened 2025-03-22
Optimize JObject Conversion

GuyPago:
Given the following example
```csharp
LanguagesHelper.TrackSet([.. raw["set"].ToArray().Select(v => $"{v}")]);
```
Without mentioning the allocations introduced by `ToArray()` and the spread operator, the underlying values are already (likely to be) strings, so the manual `v => "{v}"` conversion introduces a redundant overhead by creating extra intermediate objects.

Using `JValue.ToObject<T>` uses deserialization which uses direct conversion, and makes it a perfect choice here for both performance and error handling.

### DIFF
--- a/src/WebAPI/AdminAPI.cs
+++ b/src/WebAPI/AdminAPI.cs
@@ -241,7 +241,7 @@ public static class AdminAPI
         long lastSeq = Interlocked.Read(ref Logs.LogTracker.LastSequenceID);
         result["last_sequence_id"] = lastSeq;
         JObject messageData = [];
-        List<string> types = [.. raw["types"].Select(v => $"{v}")];
+        string[] types = raw["types"].ToObject<string[]>();
         foreach (string type in types)
         {
             Logs.LogTracker tracker;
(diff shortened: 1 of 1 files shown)


mcmonkey4eva:
While not necessarily a wrong change in and of itself - this is not performance sensitive code, `Debug` code is almost never used, `LogMessages` is used but infrequently, and that line handles a small chunk of data outside of any loop.

The usage of `.Select(v => $"{v}")` was chosen to be semantically infallible: that is to say, there's no possible confusion/mistake/bad input. It will work great on mildly bad input, and on very bad input it will give a simple and predictable error message. Someone inexperienced looking at this code can tell exactly what it's going to do and how it's going to do it.
It doesn't look clever/perfect at a glance, but it was written that way intentionally.

`ToObject<string[]>()` relies on behavior of the json.net library. Glancing at the relevant source code, I can't even figure out what route that goes down - it relies on `string[]` type conversion, with is not in the direct conversion code, but goes through a 12-step-abstraction-layer - proxies and contracts and etc and etc. to eventually deserialize somewhere. (for `string` inputs it just does a direct cast, not a proper conversion at all, but that's not applicable when it goes down the entirely separate conversion route).
It's not a pretty system and it's extremely hard to make any guarantees about how it will behave. It's also hard to have expectations that how it handles it now will be the same as how it handles it in the future, considering the `Deserialize` method has native impls for some type codes (like `string`) but abstracts away others (like `string[]`) to a separate serialization engine. If that serialization engine does have handling for not-strings as inputs currently... the native impl for `string` doesn't, so if a `string[]` native impl is added it likely won't either.

PR #706, opened 2025-04-12
feat: mount custom_nodes and models folders into SwarmUI container

gulgiraf:
Added Docker volume mounts to map local ComfyUI folders into the SwarmUI container:
- custom_nodes → /SwarmUI/dlbackend/ComfyUI/custom_nodes
- models → /SwarmUI/dlbackend/ComfyUI/models

This ensures that custom nodes and model files are available inside the container for use by the ComfyUI backend.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ Properties/
 /Output
 /Data
 /Models
+/ComfyUI
 tmp/
 /src/Extensions
 /dlbackend
--- a/launchtools/launch-standard-docker.sh
+++ b/launchtools/launch-standard-docker.sh
@@ -29,6 +29,8 @@ docker run -it \
     -v "$PWD/Models:/SwarmUI/Models" \
     -v "$PWD/Output:/SwarmUI/Output" \
     -v "$PWD/src/BuiltinExtensions/ComfyUIBackend/CustomWorkflows:/SwarmUI/src/BuiltinExtensions/ComfyUIBackend/CustomWorkflows" \
+    -v "$PWD/ComfyUI/custom_nodes:/SwarmUI/dlbackend/ComfyUI/custom_nodes" \
+    -v "$PWD/ComfyUI/models:/SwarmUI/dlbackend/ComfyUI/models" \
     --gpus=all -p 7801:7801 swarmui $POSTARG
 
 if [ $? == 42 ]; then


launchtools/launch-standard-docker.sh
```diff
@@ -29,6 +29,8 @@ docker run -it \
     -v "$PWD/Models:/SwarmUI/Models" \
     -v "$PWD/Output:/SwarmUI/Output" \
     -v "$PWD/src/BuiltinExtensions/ComfyUIBackend/CustomWorkflows:/SwarmUI/src/BuiltinExtensions/ComfyUIBackend/CustomWorkflows" \
+    -v "$PWD/ComfyUI/custom_nodes:/SwarmUI/dlbackend/ComfyUI/custom_nodes" \
```
mcmonkey4eva, line 32: This path is wrong

.gitignore
```diff
@@ -15,6 +15,7 @@ Properties/
 /Output
 /Data
 /Models
+/ComfyUI
```
mcmonkey4eva, line 18: you should not have comfy in your root

launchtools/launch-standard-docker.sh
```diff
@@ -29,6 +29,8 @@ docker run -it \
...
+    -v "$PWD/ComfyUI/models:/SwarmUI/dlbackend/ComfyUI/models" \
```
mcmonkey4eva, line 33: nothing should be in the comfy model dir

gulgiraf: @mcmonkey4eva Thanks for all your great work on the project!

Just to clarify, I’m not mounting anything from the actual ComfyUI repo — I just use ComfyUI/models as a folder name to keep things organized.

I’m not sure I see the issue with mounting to /SwarmUI/dlbackend/ComfyUI/models — could you clarify why that’s a problem or suggest a better place (or method) for model files and custom_nodes?

.gitignore
gulgiraf: @mcmonkey4eva The ComfyUI repo/system isn’t in the root directory. I created the folder to keep models and custom nodes separate from SwarmUI’s data and models. This way, both sets of folders remain accessible in the same way as SwarmUI’s.

launchtools/launch-standard-docker.sh
gulgiraf: @mcmonkey4eva In what way? I have tested the path, and it is technically correct (its working).

launchtools/launch-standard-docker.sh
mcmonkey4eva: So this PR is to add your personal folder configs to the global default docker setup? That makes no sense, just make your own copy of the script? You don't need to PR to master for a personal file

gulgiraf: @mcmonkey4eva 
This PR isn’t about personal configs. It’s about giving ComfyUI the same folder availability that SwarmUI already has, which currently only works because of how the Dockerfile is set up.

This just makes things consistent and flexible for both systems without changing default behavior. Totally fine to keep it in a fork if needed, but thought others might find it useful too.

mcmonkey4eva: No fork even... just, keep a local file for yourself.

PR #737, opened 2025-04-25
feat: add option to disable automatic hover tooltips

zerobutsugar:


mcmonkey4eva:
... Huh? What are you even trying to block here? The only hover text in the default UI is `title` attribs

zerobutsugar:
Hi! For example if you hover with the mouse over the loras in the model browser it pops up one of those tooltips and sadly the available option (for the question mark buttons) don't apply to this.

The issue is, I can't have my mouse in there, so if the lora has one of the genius long trigger phrases, I'll get a HUGE overlay with the entire text in the description and it makes it difficult to copy triggers and or scroll through, or any kind of interaction where my mouse is somewhere in there.

The same issue (even though not as annoying I'll admit) happens on the main prompt, so I thought an option would be good. I'm certain there might be a better way to handle this but for me this worked. And while I can see the benefit of having it on maybe, do not have the description in the small window I thought making it optional could be good.
Using Firefox here.

mcmonkey4eva:
Those are `title` attribs, and the majority of code in this PR is... I have no idea what it's doing, but entirely unrelated changes.

mcmonkey4eva:
I pushed a commit which removes the `title` from models if you're in a `Cards` format (ie where the info is already on screen and the title is redundant), which should fix your primary issue there.

zerobutsugar:
This PR adds the user setting to disable all automatic hover popups, including both native browser tooltips and bootstraps tooltips as well as CSS to hide bootstrap elements, js for native tooltips and apply the css. :)
Like I said, there might be a better / cleaner way for this but I thought the chances would be higher to get this implemented with a PR and did not want to rip it out like a maniac.

zerobutsugar:
Thanks, I'll give it a try!

mcmonkey4eva:
I don't think bootstrap tooltips exist anywhere in Swarm?

Swarm's popovers, I see no reason to disable at all, they're already very user-controllable.

zerobutsugar:
Thanks! working as intended.
When i started on this I thought because of the hover triggered tooltips that it could use elements of bootstrap, so in the end it was just a proactive measure to cover any possible scenario of any tooltips. so I'll take it, it was a bit like shooting sparrows with a cannon but happy its implemented now :)

PR #738, opened 2025-04-25
Linux Per Distro Install.md

Eugeniusz-Gienek:
Simple as it is. Manuals - how to install in different Linux distros. Currently only Gentoo, later more.


mcmonkey4eva:
what... the heck is that. No.

Eugeniusz-Gienek:
Umm any comments on that?..

Eugeniusz-Gienek:
Rude...

PR #742, opened 2025-04-26
Update extension_list.fds

Hugs288:


mcmonkey4eva:
I'm confused

https://github.com/Hugs288/NegPipforSwarmUI/blob/main/NegPipforSwarmUI.cs#L51

You're adding the node... before absolutely anything else at all ever? That doesn't seem like a good idea. Especially the odd choice of ultra low negative value

mcmonkey4eva:
Also, is an entire param group needed for a single parameter? 0.o

Hugs288:
oop, good catch i did that for testing purposes, ill make it -10, i think thats fine?

as for the param group, idk a better way to do it, sorry.

mcmonkey4eva:
`-10` is when loras load, most guidance-muckery-type-stuff is `-7`

and: you can put it into any existing group that fits appropriately

Hugs288:
done, put it under the sampling group.

mcmonkey4eva:
You'll also want `IgnoreIf: "false"` so it doesn't add to metadata when disabled

mcmonkey4eva:
and to remove the group that does nothing

mcmonkey4eva:
aaand update the JS asset to put the installer in the Sampling group that's in use now instead of the old group

PR #748, opened 2025-04-30
Added sampler and scheduler option to refiner

### LINKED ISSUE #611: Different sampler/cfg/scheduler within refiner / hires "fix"
### Feature Idea

Within refiner / upscale part, add a hidden[?] set of options to set Refiner Sampler, Refiner Scheduler, Refiner CFG, etc; The base option for dropdowns should be "Same as Base", which just uses the base sampler sched etc.

### Other

something something cfgpp doesnt like upscaling

akadroid:
Fixes #611 
It uses the base sampler and scheduler when toggled off.

### DIFF
--- a/src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
+++ b/src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
@@ -1546,8 +1546,8 @@ public class WorkflowGenerator
             ["steps"] = steps,
             ["cfg"] = cfg,
             // TODO: proper sampler input, and intelligent default scheduler per sampler
-            ["sampler_name"] = UserInput.Get(ComfyUIBackendExtension.SamplerParam, defsampler ?? DefaultSampler),
-            ["scheduler"] = UserInput.Get(ComfyUIBackendExtension.SchedulerParam, defscheduler ?? DefaultScheduler),
+            ["sampler_name"] = defsampler ?? UserInput.Get(ComfyUIBackendExtension.SamplerParam, defsampler ?? DefaultSampler),
+            ["scheduler"] = defscheduler ?? UserInput.Get(ComfyUIBackendExtension.SchedulerParam, defscheduler ?? DefaultScheduler),
             ["positive"] = pos,
             ["negative"] = neg,
             ["latent_image"] = latent,
--- a/src/Text2Image/T2IParamTypes.cs
+++ b/src/Text2Image/T2IParamTypes.cs
@@ -2,6 +2,7 @@
 using FreneticUtilities.FreneticToolkit;
 using Newtonsoft.Json.Linq;
 using SwarmUI.Accounts;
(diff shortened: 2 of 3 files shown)


src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
```diff
@@ -1546,8 +1546,8 @@ static bool isSpecial(T2IModel model)
             ["steps"] = steps,
             ["cfg"] = cfg,
             // TODO: proper sampler input, and intelligent default scheduler per sampler
-            ["sampler_name"] = UserInput.Get(ComfyUIBackendExtension.SamplerParam, defsampler ?? DefaultSampler),
-            ["scheduler"] = UserInput.Get(ComfyUIBackendExtension.SchedulerParam, defscheduler ?? DefaultScheduler),
+            ["sampler_name"] = defsampler ?? UserInput.Get(ComfyUIBackendExtension.SamplerParam, defsampler ?? DefaultSampler),
```
mcmonkey4eva, line 1549: this ain't right, this just disables the sampler param

mcmonkey4eva:
The reason that issue was hesitated on is just adding new params is pretty dirty at this point, the refine/upscale section is too big - the ideal solution is probably to implement sub-grouping, so that "override params" like cfg/steps/sampler/etc can be in a subgroup rather than littering the main group even more

src/Text2Image/T2IParamTypes.cs
```diff
@@ -486,6 +487,14 @@ static List<string> listRefinerModels(Session s)
             + "Too-high values can cause corrupted/burnt images, too-low can cause nonsensical images.\n7 is a good baseline. Normal usages vary between 4 and 9.\nSome model types, such as Turbo, expect CFG around 1.",
             "7", Min: 0, Max: 100, ViewMax: 20, Step: 0.5, Examples: ["5", "6", "7", "8", "9"], OrderPriority: -3.5, ViewType: ParamViewType.SLIDER, Group: GroupRefiners, ChangeWeight: -3, Toggleable: true, IsAdvanced: true
             ));
+        RefinerSampler = Register<string>(new("Refiner Sampler", "Alternative sampler for the refiner stage. When enabled, this sampler will replace the base sampler for the refiner stage.",
+            "euler", Toggleable: true, FeatureFlag: "refiners", Group: GroupRefiners, OrderPriority: -3.25, IsAdvanced: true,
+            GetValues: (_) => ComfyUIBackendExtension.Samplers
```
mcmonkey4eva, line 492: this is the wrong place to register this param

mcmonkey4eva:
As a new contributor please select from the "Easy PR" tag for your first PR(s), to get familiar with the codebase before you take on issues with hidden complexity

akadroid:
Damn alright
I couldn't get it to use the defsampler since it kept getting overridden by the samplerparam so I had to implement it that way. I would love to see this get implemented properly in the future

PR #764, opened 2025-05-06
standardize seed controls

Hugs288:
this pr does 3 things:

1) wildcard seed uses the same seed as the image gen seed if not specified, instead of image gen seed + 17 (why?)

2) the reuse parameters button only reuses seed values if they are not the same as the image gen seed (works fine for extensions too as long as the seed metadata includes "seed").

3) remove the wildcardseed exception from the reuse params button, as it is no longer necessary.

the main reason for this pr is that i found it pretty annoying having to reset tipo seed after every time i reuse params from an image generated with tipo, but i didnt want to exclude it either. and less finnicky wildcard/extension seeds is nice too i think.

but it is a pretty subjective pr, i understand if you dont want it merged for whatever reason.

i have also updated [tipoforswarmui](https://github.com/Hugs288/TipoForSwarmUI) to take advantage of these changes.

### DIFF
--- a/src/Core/Settings.cs
+++ b/src/Core/Settings.cs
@@ -466,7 +466,7 @@ public class Settings : AutoConfiguration
         public string Language = "en";
 
         [ConfigComment("Comma-separated list of parameters to exclude from 'Reuse Parameters'.\nFor example, set 'model' to not copy the model, or 'model,refinermodel,videomodel' to really never copy any models.")]
-        public string ReuseParamExcludeList = "wildcardseed";
+        public string ReuseParamExcludeList = "";
 
         /// <summary>Settings related to audio.</summary>
         public class AudioData : AutoConfiguration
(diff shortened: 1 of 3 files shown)


mcmonkey4eva:
`wildcard seed uses the same seed as the image gen seed if not specified, instead of image gen seed + 17 (why?)`
All seeds are intentionally internally offset different amounts to avoid stacking errors. This is most noticeable with latent seed (eg the image model seed and the refiner seed are offset - without that offset it produces immediate visual distortion in the image), but applies to more simple seeding like the C# random number generator as well.

I pushed a commit that instead adds the offset to both the manual input and the random seed, so that you don't get surprised by different results with manual vs base seed.

---------

`the reuse parameters button only reuses seed values if they are not the same as the image gen seed (works fine for extensions too as long as the seed metadata includes "seed").`
That is hacky? iirc there's a metadata trick somewhere that applies as a more proper fix.

--------------

`remove the wildcardseed exception from the reuse params button, as it is no longer necessary.`
no

------------

`i have also updated [tipoforswarmui](https://github.com/Hugs288/TipoForSwarmUI) to take advantage of these changes`
I'd recommend adding a unique offset there too

------------

I appreciate you asking `why?`, but in the future I recommend asking that *before* changing things -- [_Chesterton's Fence_](https://fs.blog/chestertons-fence/) principle

PR #765, opened 2025-05-07
Update README.md

Systho:
fix typo in Readme for downloading the same version as current tag

you might want to cherry-pick and merge this change on the 0.9.6-Beta tag as well, but I'm not sure what is your policy about this and the change is arguably without any value. 
Changing on master is more important since this is the file displayed when navigating to the repo.


Systho:
The release does not have that file, my bad.

PR #771, opened 2025-05-11
add image quality setting, remove redundant presets

Hugs288:


src/Utils/Image.cs
```diff
@@ -247,18 +239,14 @@ public static string ImageFormatToExtension(string format)
         {
             "PNG" => "png",
             "JPG" => "jpg",
-            "JPG90" => "jpg",
-            "JPG75" => "jpg",
             "WEBP_LOSSLESS" => "webp",
-            "WEBP_100" => "webp",
-            "WEBP_90" => "webp",
-            "WEBP_75" => "webp",
```
mcmonkey4eva, line 255: all the legacy inputs need to stay present for compat with older data

src/Utils/Image.cs
```diff
@@ -301,29 +289,15 @@ public Image ConvertTo(string format, string metadata = null, int dpi = 0)
                 ext = "png";
                 break;
             case "JPG":
-                img.SaveAsJpeg(ms, new JpegEncoder() { Quality = 100 });
-                break;
-            case "JPG90":
-                img.SaveAsJpeg(ms, new JpegEncoder() { Quality = 90 });
```
mcmonkey4eva, line 307: same here as above - anyone who had JPG90 selected needs to not have their swarm suddenly break because of this change.

Hugs288:
done, also changed the default to 100 so jpg behaves the same

PR #774, opened 2025-05-14
Outline currently selected image

CreativeAlloy:
This new logic adds an outline to the currently selected image, whether from the image history or the current image batch.

More specifically, I have modified three files with the following code:
- genpage.css, which now contains style code for .image-block-selected, the outline using var --emphasis' color code to ensure that it's consistent with the current loaded theme.
- currentimagehandler.js, which now contains logic within function clickImageInBatch(div) that, upon clicking any image, removes the outline from any previously selected image, and adds the latest clicked image to the classList instead. Furthermore, some additional logic uses the existing "curImg.dataset.batch_id = 'history';" code to make an if(){} check, where if the batchId **is NOT** history, the selector will be removed from the image history as well.
- imagehistory.js, with similar logic to remove and add images into the classList.

There's also a few line breaks removed before some "else"s, which I left as they were. I can easily go back and reinsert the line breaks if desired.

Attached example images to showcase what this looks like with the default dark theme.

image
image

### DIFF
--- a/src/wwwroot/js/genpage/gentab/currentimagehandler.js
+++ b/src/wwwroot/js/genpage/gentab/currentimagehandler.js
@@ -237,6 +237,13 @@ function toggleShowLoadSpinners() {
 }
 
 function clickImageInBatch(div) {
+    // Remove 'image-block-selected' from all image-blocks in the batch
+    document.querySelectorAll('#current_image_batch .image-block-selected').forEach(function (block) {
+        block.classList.remove('image-block-selected');
+    });
+    // Add 'image-block-selected' to the clicked block
+    div.classList.add('image-block-selected');
+
     let imgElem = div.getElementsByTagName('img')[0];
     if (currentImgSrc == div.dataset.src) {
         imageFullView.showImage(div.dataset.src, div.dataset.metadata);
@@ -244,6 +251,7 @@ function clickImageInBatch(div) {
     }
     setCurrentImage(div.dataset.src, div.dataset.metadata, div.dataset.batch_id ?? '', imgElem && imgElem.dataset.previewGrow == 'true', false, true, div.dataset.is_placeholder == 'true');
 }
+
 
 function rightClickImageInBatch(e, div) {
     if (e.shiftKey || e.ctrlKey) {
@@ -369,8 +377,7 @@ function shiftToNextImagePreview(next = true, expand = false) {
         let newIndex = index + (next ? 1 : -1);
         if (newIndex < 0) {
             newIndex = divs.length - 1;
-        }
-        else if (newIndex >= divs.length) {
+        } else if (newIndex >= divs.length) {
             newIndex = 0;
         }
         divs[newIndex].querySelector('img').click();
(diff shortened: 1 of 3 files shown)


src/wwwroot/js/genpage/gentab/currentimagehandler.js
```diff
@@ -237,6 +237,13 @@ function toggleShowLoadSpinners() {
 }
 
 function clickImageInBatch(div) {
+    // Remove 'image-block-selected' from all image-blocks in the batch
+    document.querySelectorAll('#current_image_batch .image-block-selected').forEach(function (block) {
```
mcmonkey4eva, line 241: Please use standard `foreach` not functional forEach.
ie, `foreach (let block of document.queryetc)`

src/wwwroot/js/genpage/gentab/currentimagehandler.js
```diff
@@ -237,6 +237,13 @@ function toggleShowLoadSpinners() {
 }
 
 function clickImageInBatch(div) {
+    // Remove 'image-block-selected' from all image-blocks in the batch
```
mcmonkey4eva, line 240: these comments are redundant

src/wwwroot/js/genpage/gentab/currentimagehandler.js
```diff
@@ -245,6 +252,7 @@ function clickImageInBatch(div) {
     setCurrentImage(div.dataset.src, div.dataset.metadata, div.dataset.batch_id ?? '', imgElem && imgElem.dataset.previewGrow == 'true', false, true, div.dataset.is_placeholder == 'true');
 }
 
+
```
mcmonkey4eva, line 255: minor stray edit

src/wwwroot/js/genpage/gentab/currentimagehandler.js
```diff
@@ -369,8 +377,7 @@ function shiftToNextImagePreview(next = true, expand = false) {
         let newIndex = index + (next ? 1 : -1);
         if (newIndex < 0) {
             newIndex = divs.length - 1;
-        }
-        else if (newIndex >= divs.length) {
+        } else if (newIndex >= divs.length) {
```
mcmonkey4eva, line 380: no put those back

src/wwwroot/js/genpage/gentab/currentimagehandler.js
mcmonkey4eva: Also this should probably be current_image_batch querySelectorAll, not document query that then only matches within current_image_batch

src/wwwroot/js/genpage/gentab/currentimagehandler.js
```diff
@@ -392,12 +395,15 @@ function shiftToNextImagePreview(next = true, expand = false) {
...
     if (newIndex < 0) {
         newIndex = imgs.length - 1;
-    }
-    else if (newIndex >= imgs.length) {
+    } else if (newIndex >= imgs.length) {
         newIndex = 0;
     }
     let newImg = imgs[newIndex];
     let block = findParentOfClass(newImg, 'image-block');
+    for (let block of getRequiredElementById('current_image_batch').querySelectorAll('.image-block-selected')) {
```
mcmonkey4eva, line 403: I think setting the selection here is redundant if `setCurrentImage` is going to replace it right after anyway?

src/wwwroot/js/genpage/gentab/imagehistory.js
```diff
@@ -139,6 +139,10 @@ function describeImage(image) {
 function selectImageInHistory(image, div) {
     lastHistoryImage = image.data.src;
     lastHistoryImageDiv = div;
+    for (let selected of document.querySelectorAll('.browser-list-entry.image-block-selected, .browser-details-list-entry.image-block-selected, .image-block.image-block-selected, .model-block.image-block-selected')) {
```
mcmonkey4eva, line 142: Same as other comment, this will be overwritten by `setCurrentImage`?

mcmonkey4eva:
Been a few months without update, gonna close for now. Feel free to resubmit later.

Issue #713, opened 2025-04-17
Allow easy X out of filter search in bottom tabs like in the parameters tab


bouchon1996:
### Feature Idea

Searching in the parameters currently lets you easily X out of a search whereas searches on the bottom tabs such as LORAs, VAEs, etc. do not have the X. It would be useful to use this same easy X out wherever a filter search is located.

See the example image for a comparison of the different search boxes:

![Image](https://github.com/user-attachments/assets/44d187b7-363d-46c2-88e4-514b038610d7)

halfblack:
I came here to request this exact feature.

mcmonkey4eva:
Added thanks to @Jelosus2 !

PR #776, opened 2025-05-15
Add clear button (X) to the filter inputs at the bottom

Jelosus2:
PR to implement the feature of the issue #713


mcmonkey4eva:
Thank you!

Also in the future ftr when github asks whether to allow edits from maintainers, that should be enabled.

PR #789, opened 2025-05-23
fix webp metadata

Hugs288:
fixes encoding metadata into webp images

however dragging a webp image into swarm still doesnt show its metadata, i have no idea why, sorry

images from image history show metadata fine tho


Hugs288:
ive tested all formats with this change and they work fine

Issue #798, opened 2025-05-29
Flux fill doesn't work as a segment model


willhsmit:
### Expected Behavior

I should be able to use Flux fill as the model for refining a `<segment>` tag, either with
(i) using Flux fill as the base model for an inpaint, leaving the Segment Model unset, and using the `<Segment>` tag to further refine part of the inpaint with Flux fill
(ii) Using Flux dev as the base model for a generation, setting the Segment Model to flux fill, and using the `<Segment>` tag to further refine part of the gen with Flux fill.

(Both of these use cases fail for what I think are related reasons)

### Actual Behavior

In case (i) (FluxFill as both base and segment model), the segment attempts to redraw a miniature version of the entire input image in the segment area.

You can see in the following image that the entire forest - not just the segment or the masked area - is redrawn over the woman's face (prompt was: a blond woman smiling at the camera, wearing a green shirt, waist up medium shot `<segment:face,0.9,0.5>` closeup of a young blond woman's face, smiling)

![Image](https://github.com/user-attachments/assets/caef6136-7bf9-43d0-ab78-910e4ba0eb5f)

In case (ii), (FluxDev as base model and FluxFill as segment model), the gen fails altogether with:

17:10:33.972 [Warning] [ComfyUI-0/STDERR]   File "C:\Users\willh\SwarmUI\dlbackend\comfy\ComfyUI\execution.py", line 349, in execute
17:10:33.973 [Warning] [ComfyUI-0/STDERR]     output_data, output_ui, has_subgraph = get_output_data(obj, input_data_all, execution_block_cb=execution_block_cb, pre_execute_cb=pre_execute_cb)
17:10:33.973 [Warning] [ComfyUI-0/STDERR]                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
17:10:33.974 [Warning] [ComfyUI-0/STDERR]   File "C:\Users\willh\SwarmUI\dlbackend\comfy\ComfyUI\execution.py", line 224, in get_output_data
17:10:33.975 [Warning] [ComfyUI-0/STDERR]     return_values = _map_node_over_list(obj, input_data_all, obj.FUNCTION, allow_interrupt=True, execution_block_cb=execution_block_cb, pre_execute_cb=pre_execute_cb)
[...]

PR #803, opened 2025-06-01
Fix SwarmUI issue #798 by sending pixel/mask info into CreateKSampler from the segment call site.

willhsmit:
When using flux fill as the model for segment refining, with either Flux fill or dev as the original model, InpaintModelConditioning needs the masked output pixels and mask from the segment clip/crop nodes, not just the latent. The existing code always uses the original input pixels and mask. (See examples in https://github.com/mcmonkeyprojects/SwarmUI/issues/798).

This PR adds parameters to CreateKSampler to pass these in, and sets the parameters in the callsite for the Segment flow, where they're readily available. 

Caveats:
* Only the segment call to CreateKSampler is updated; I don't believe any of the other callsites are good candidates for this but I might be wrong. Several of them are video-specific; currently upscale/refiner refines the whole image and doesn't use a mask, so it's not a good candidate for Flux fill. 
* Using a different Flux model for the segment pass than the original pass (dev for a T2I, fill for a segment; Fill for a big inpaint, dev for a refine) works without an Error now, but quality is still limited by the fact that Fill and Dev want very different Flux Guidance settings, and there isn't a separate setting for the Segment Model flux guidance.

This is the new output for the example image from #798 

1134-a blond woman smiling at the camera, wea-flux-fill-nunchaku-int4transformer_bloc-552300383

And the exported Comfy workflow (hard to screenshot) shows the nodes in a more logical sequence.

[FluxSegmentUpdated.json](https://github.com/user-attachments/files/20542012/FluxSegmentUpdated.json)


mcmonkey4eva:
Valid workflow detail analysis, but the code approach is a bit off - I've pushed a commit to fix it

willhsmit:
Thanks!

PR #807, opened 2025-06-04
Reorder prompt elements using alt-left and alt-right

ribawaja:
Reorder prompt elements using alt + left/right arrow.
Alt-left moves the element under the cursor to the left. Alt-right behaves similarly.
Elements are assumed to be comma-separated.

Inspired by this a1111 feature:
https://github.com/AUTOMATIC1111/stable-diffusion-webui/issues/10593

### DIFF
--- a/src/wwwroot/js/site.js
+++ b/src/wwwroot/js/site.js
@@ -344,9 +344,73 @@ function textPromptAddKeydownHandler(elem) {
         }
         triggerChangeFor(elem);
     }
+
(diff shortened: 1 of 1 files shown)


mcmonkey4eva:
The LLM that wrote this PR did a pretty bad job of basic syntax, but managed a good enough core concept that I just rewrote it to be less stupid

mcmonkey4eva:
This behavior was altered to be behind user setting `TagMoveHotkeyEnabled`

PR #833, opened 2025-06-18
Add UI elements and logic to let users sort the presets in the preset view

bman654:
Works similar to the way model sorting works.  defaults to the existing sort order (which is something similar to but not always "last modified").

### DIFF
--- a/src/wwwroot/js/genpage/gentab/presets.js
+++ b/src/wwwroot/js/genpage/gentab/presets.js
@@ -1,5 +1,6 @@
 
 let allPresets = [];
+let allPresetsUnsorted = [];
 let currentPresets = [];
 
 let preset_to_edit = null;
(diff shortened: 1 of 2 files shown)


src/wwwroot/js/genpage/gentab/presets.js
```diff
@@ -292,16 +293,91 @@ function editPreset(preset) {
...
 }
 
+let presetSortByElement;
+let presetSortReverseElement;
+let presetSortBy = localStorage.getItem('preset_list_sort_by') || 'Default';
+let presetSortReverse = localStorage.getItem('preset_list_sort_reverse') === 'true';
+
+
+function onPresetSortByChanged(ev) {
+    localStorage.setItem('preset_list_sort_by', presetSortBy = ev.target.value);
```
mcmonkey4eva: there's a lot of odd code format choices in here

src/wwwroot/js/genpage/gentab/presets.js
```diff
@@ -292,16 +293,91 @@ function editPreset(preset) {
...
+    localStorage.setItem('preset_list_sort_reverse', presetSortReverse = ev.target.checked);
+    presetBrowser.update();
+}
+
+function getSortInfo() {
+    const sortByElem = document.getElementById("preset_list_sort_by");
+    const reverseElem = document.getElementById("preset_list_sort_reverse");
+
+    if (sortByElem !== presetSortByElement) {
+        presetSortByElement?.removeEventListener("change", onPresetSortByChanged);
```
mcmonkey4eva: whaaaat is going on here

src/wwwroot/js/genpage/gentab/presets.js
```diff
@@ -292,16 +293,91 @@ function editPreset(preset) {
...
+        }
+    }
+    
+    return {sortBy: presetSortBy, reverse: presetSortReverse};
+}
+
+function getSortCompareFunc(sortBy, reverse) {
+    const getValue = preset => {
+        if (preset.title.toLowerCase() === "default") {
+            return "          default";
```
mcmonkey4eva: odd hack

src/wwwroot/js/genpage/gentab/presets.js
```diff
@@ -292,16 +293,91 @@ function editPreset(preset) {
...
+        if (preset.title.toLowerCase() === "preview") {
+            return "          preview";
+        }
+        
+        switch (sortBy) {
+            // for Name, use title, but strip off the path part
+            case 'Name': return preset.title.substring(preset.title.lastIndexOf('/') + 1);
+            case 'Title': return preset.title;
+            case 'Default':
+            default: return 'z'; // stable sort will basically do nothing
```
mcmonkey4eva: if no sorting is requested, it should just not call a sort func

mcmonkey4eva:
This is a really odd code approach. Why is it not just identical to image history or model sort code?

src/wwwroot/js/genpage/gentab/presets.js
```diff
@@ -292,16 +293,91 @@ function editPreset(preset) {
...
+            return "          default";
+        }
+        if (preset.title.toLowerCase() === "preview") {
+            return "          preview";
+        }
+        
+        switch (sortBy) {
+            // for Name, use title, but strip off the path part
+            case 'Name': return preset.title.substring(preset.title.lastIndexOf('/') + 1);
+            case 'Title': return preset.title;
```
mcmonkey4eva: 'name' / 'title' split here is odd

src/wwwroot/js/genpage/gentab/presets.js
bman654: can you provide any specifics on what you find odd about the format?

src/wwwroot/js/genpage/gentab/presets.js
bman654: I was running into issues with the actual lifetime of the html elements vs when this code executes.  Finally wrote it to defensively check the element each time and add/remove listeners as needed.

src/wwwroot/js/genpage/gentab/presets.js
bman654: I can remove it.

src/wwwroot/js/genpage/gentab/presets.js
bman654: I should rename "Title" -> "Path" to hopefully be more clear that one sorts by just the basename and one by the full path (which is called Title in the UI).

src/wwwroot/js/genpage/gentab/presets.js
mcmonkey4eva: This commented line has snuck an inline value assignment in the middle of another line. There's double-blanklines all around. There's usage of syntax features that the rest of the codebase generally doesn't use, eg `const` and `===`
The most important thing code in a PR needs to be, syntax wise, is similar to the other code around it in the same project. Otherwise it becomes an issue down the road for maintenance.

bman654:
> This is a really odd code approach. Why is it not just identical to image history or model sort code?

Both Model and Image History use server-side sorting and have an API endpoint dedicated to retrieving them in sorted order.

Presets does not use server-side sorting and also does not have a dedicated endpoint (looks like they come over as part of "UserData").

I'm working with what's available.

src/wwwroot/js/genpage/gentab/presets.js
mcmonkey4eva: You can just not store it then, and use `getRequiredElementById`

src/wwwroot/js/genpage/gentab/presets.js
bman654: ok I've gone through and cleaned up the spacing, added comments, and (even though it feels really weird) removed the usages of `const` and `===` in favor of the looser `let` and `==` syntax.

Let me know if it's still out of line with your JS usage preferences.

src/wwwroot/js/genpage/gentab/presets.js
bman654: but that function will throw if the element does not exist (which it sometimes doesn't when this code runs). Errors about the element not existing is what led me to wrap it in this code.

I guess since you aren't using any specific front end framework I don't have any real understanding of the lifecycle of the UI elements or order of operations so I'm falling back on patterns that I know are robust when coding straight against the DOM.

Your existing code (models for example) also had logic to deal with this problem (code running before elements exist) but it was coupled to re-fetching data from the server.  For example, some of the logic about adding event listeners was tied to the value of `isRefresh` parameter.  I don't have enough context about when or why the code was being called with specific values of `isRefresh` to feel confident I could lift the same logic and apply it to presets (which have some differences in how they are fetched compared to models).

mcmonkey4eva: I again say go look at the image history code, which already handles this stuff.

If you make up a whole new way to solve a problem, that means if something bugs... there's now two different implementations that need to be separately looked into and separately fixed. If there's a feature to add, there's two different implementations that need to have that separately added to.

As much as possible, new things should either use, or exactly copy, the logic of pre-existing code. Not because one way is necessarily better than the other, but because it makes long term maintenance a lot easier.

bman654: I've done my best to make it similar to image history.  But the lifecycles are different and it can't be the same.  Presets get initialized differently than image history and as a result the wiring up of the event listeners just can't be done the same way.  Hopefully this is close enough to what you are looking for.

PR #850, opened 2025-06-27
Add <setvarq> prompt syntax for setting variables without emitting to the prompt

bman654:


mcmonkey4eva:
I implemented this in a simpler way - you can now eg `<setvar[name,false]:...>` to indicate that the set should not emit, allowing also for easy consistency with `setmacro` to work the same way

bman654:
> I implemented this in a simpler way - you can now eg `<setvar[name,false]:...>` to indicate that the set should not emit, allowing also for easy consistency with `setmacro` to work the same way

FWIW I thought about the ,false way, but didn't do it b/c AFAICT that is a breaking change since the original code put almost no constraints on the format of variable names.

PR #854, opened 2025-06-29
Expose SamplerSigmaMax for all model types

bWm-nubby:
Remove SD3 model type restriction for displaying Sigma Max value in Advanced Sampling.


mcmonkey4eva:
I. have no idea why that was there. weird

PR #859, opened 2025-07-02
Add user setting to control fields shown in Preset list.

bman654:
image

image

image

image

### DIFF
--- a/languages/en.json
+++ b/languages/en.json
@@ -396,8 +396,6 @@
         "PortCanChange": "",
         "Prefix a line with # to make it a comment (ie won't be counted as an option).": "",
         "Preset description text...": "",
-        "PresetListDisplayFields": "",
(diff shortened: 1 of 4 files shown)


mcmonkey4eva:
Merged in the edits intended to the Details display, changes outside of that view mode don't quite make sense

bman654:
Just to make the case for the other view modes....I posted above pictures of what it does to each view mode.
Compare that against this that we get currently for Cards:
image

This view is useless to me b/c the limited real estate is taken up by the prompt (which I do not find useful to see in this view) and thus the fields that are important to me (Steps, CFG, Sampler, ...) are hidden.

For Thumbnails, the tooltip situation is not quite as bad.  This is what we get currently:

image

I can see everything in it.  But I'd like to make the argument that the version in this PR which honors the param priorities the user specifies is more legible - it places what the user considers the most important params up front and puts the remainder below the blank line and this makes it easier to process whatever the user considers most important.

PR #869, opened 2025-07-05
add png stealth info support

Hugs288:
adds stealth png info support

this implementation only works with png, it is compatible with webp in theory but either sixlabors doesnt support that or stupid me just cant get it to work.

especially tricky for lossy webp since it needs the alpha layer to be encoded losslessly, which sixlabors doesnt seem to support at all.

also small change in imagemetadatatracker because webp metadata seems fixed now


mcmonkey4eva:
This is pretty cool, but please reset your git and resubmit the code - this PR has 30 random commits

In the future, always make sure to create a branch before making edits, otherwise you'll end up with a polluted master branch

ghost:
fair

PR #871, opened 2025-07-06
stealth png info (again)

Hugs288:
adds stealth png info support

this implementation only works with png, it is compatible with webp in theory but either sixlabors doesnt support that or stupid me just cant get it to work.

especially tricky for lossy webp since it needs the alpha layer to be encoded losslessly, which sixlabors doesnt seem to support at all.

also small change in imagemetadatatracker because webp metadata seems fixed now

### DIFF
--- a/src/Utils/ImageMetadataTracker.cs
+++ b/src/Utils/ImageMetadataTracker.cs
@@ -113,7 +113,7 @@ public static class ImageMetadataTracker
     }
 
     /// <summary>File format extensions that even can have metadata on them.</summary>
-    public static HashSet<string> ExtensionsWithMetadata = ["png", "jpg"];
+    public static HashSet<string> ExtensionsWithMetadata = ["png", "jpg", "webp"];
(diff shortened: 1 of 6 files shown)


src/Utils/ImageMetadataTracker.cs
```diff
@@ -113,7 +113,7 @@ public static ImageDatabase GetDatabaseForFolder(string folder)
     }
 
     /// <summary>File format extensions that even can have metadata on them.</summary>
-    public static HashSet<string> ExtensionsWithMetadata = ["png", "jpg"];
+    public static HashSet<string> ExtensionsWithMetadata = ["png", "jpg", "webp"];
```
mcmonkey4eva: unrelated change shouldn't be here

ghost: "also small change in imagemetadatatracker because webp metadata seems fixed now"

PR #894, opened 2025-07-14
Segment union

bman654:
Before I flesh this PR out with:

* additional text in the Segment section of wiki
* additional text in the prompt autocomplete
* <stretch goal> additional UI in the modal segment wysiwyg editor

I wanted to first take your temperature on the actual feature.  It let's a user combine multiple clipseg masks and run them in a single pass through the segment model (saving imagegen time).

Example:

image

The syntax is:

```
<segment: face || hair:0.25, 0.6, 0.4>multi-colored hair, blonde hair, brown hair, long hair, ponytail, blue eyes, tears, frown
```

### DIFF
--- a/docs/Features/Prompt Syntax.md
+++ b/docs/Features/Prompt Syntax.md
@@ -151,10 +151,10 @@
             - Note the default "confidence threshold" for Yolo models is `0.25`, which is different than is often used with ClipSeg, and does not have a "max threshold" like ClipSeg does.
         - If you have a yolo model with multiple supported classes, you can filter specific classes by appending `:<classes>:` to the model name where `<classes>` is a comma-separated list of class IDs or names, e.g., `<segment:yolo-modelnamehere:0,apple,2:,0.8,0.25>`
     - You can also combine multiple areas into a single segment to refine them as a single group.
-        - Separate the areas with `||` in `texthere`
-        - For example, `<segment: face || hair>` will find all the faces and hair in the image and refine them as a single group.
+        - Separate the areas with `|` in `texthere`
+        - For example, `<segment:face|hair>` will find all the faces and hair in the image and refine them as a single group.
         - This works with YOLOv8 models as well.
-            - `<segment: yolo-face_yolov8m-seg_60.pt || yolo-hair_yolov8m-seg_60.pt || fingers>` will refine the group of faces and hair (found by YOLO) and fingers (found by CLIPSeg) as a single group.
+            - `<segment:yolo-face_yolov8m-seg_60.pt | yolo-hair_yolov8m-seg_60.pt | fingers>` will refine the group of faces and hair (found by YOLO) and fingers (found by CLIPSeg) as a single group.
     - There's an advanced parameter under `Segment Refining` named `Segment Model` to customize the base model used for segment processing
     - There's also a parameter named `Save Segment Mask` to save a preview copy of the generated mask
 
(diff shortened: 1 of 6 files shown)


launch-macos-dev.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
```
mcmonkey4eva: this file is unrelated to this PR and should not be here

mcmonkey4eva:
This is not a feature I ever even thought of, but makes sense.

I would say it should be extracted one step up the line, such that eg `<segment:face||yolo-myfacemodel.pt` would autocombine both clipseg and yolo match in one.

I don't think those mask nodes need to be used -- the ones that are intended to do regional overlapping. There's surely just a normal mask merge node somewhere

bman654:
I'm a ComfyUI noob.  From what I can tell, there's a variety of mask merging nodes, but none of them part of the default set -- they are all part of one nodepack or another.  Is there a place in SwarmUI where you define 3rd party node packs to be installed?

mcmonkey4eva:
You absolutely do not need a custom node for something so basic

mcmonkey4eva:
image

Probably `MaskComposite` in `or` mode?

bman654:
ok the UI is updated to support these changes.

Here is what the Autosegment popup looks like.  Ideally there'd be a way to add/remove masks, but I figured just putting room for 3 would cover 99.9% of the use cases.

image

Note that there is a dependency on this PR for ComfyUI: https://github.com/comfyanonymous/ComfyUI/pull/8915

Without that, attempts to combine 2 masks will fail with an error.

mcmonkey4eva:
I... think the UI should probably just be left at one mask, rather than trying to support multiple? Consider multiple an advanced case.
=or= hide the extra inputs behind a checkbox or an "add another match" button or something.
The UI screens are meant to be very simple and friendly, so avoid complexity or density where possible.

Also: ftr it's technically a mistake in any swarm nodes outputting a 2d mask - legacy code. It used to be 2d mask was normal, but in modern comfy all mask and image tensors should have a batch dimension. That comfy PR offhand looks valid regardless, but yeah we should fix the swarm mask nodes to all be batch-friendly.

bman654:
I’m inclined to just revert the change to the popup. It’s a lot of ugly code to support something that is questionable anyway given your goals.I’ll see if I can update the segment nodes without breaking anything. I’d like to remove the dependency on that other PR.  And I had no idea what the 3rd dimension was.

bman654:
ok I've:

1. reverted the segment modal to its original, more simple version
2. modified both ClipSeg and Yolo nodes to output 3d masks

Both seem to be working without the modified ComfyUI composite node now.

mcmonkey4eva:
- The inline docs were getting way too long, I swapped the last few lines to just a "if you want more info, click here" thing.
- I get what you're going for with the `;0.5` thingy, and something like that is a good idea, but that needs further thought as to how to form the syntax rather than just delving even deeper into an already pretty dirty syntax piling (there's way too many magic symbols going on), so I'm excluding it for now to pull in the main function.
- Switched to single pipe instead of double to reduce conflict risk with other list syntaxes

mcmonkey4eva:
Honestly might need to nuke half the syntax here and build a new, separate thing for "advanced segmentation" distinct from the "basic segmentation". With 'advanced' controlled directly by a UI or via a more json-esque syntax.

bman654:
expanding the Segment section of the generate params to allow defining the segments would be cool

PR #900, opened 2025-07-17
Add Image Selector in SwarmInputImage

Odls:
Add Image Selector in SwarmInputImage, just like the original LoadImage node.
It works in all three tabs: ComfyUI, Generate, and Simple

SwarmInputImageImageSelector

### DIFF
--- a/src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmInputNodes.py
+++ b/src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmInputNodes.py
@@ -1,6 +1,7 @@
 from . import SwarmLoadImageB64
 import folder_paths
-from nodes import CheckpointLoaderSimple
+from nodes import CheckpointLoaderSimple, LoadImage
+import os
 
 INT_MAX = 0xffffffffffffffff
 INT_MIN = -INT_MAX
@@ -187,15 +188,21 @@ class SwarmInputBoolean:
         return (value, )
 
 
-class SwarmInputImage:
+class SwarmInputImage(LoadImage):
     @classmethod
     def INPUT_TYPES(s):
+        input_dir = folder_paths.get_input_directory()
+        files = [f for f in os.listdir(input_dir) if os.path.isfile(os.path.join(input_dir, f))]
+        files = folder_paths.filter_files_content_types(files, ["image"])
         return {
             "required": {
-                "title": ("STRING", {"default": "My Image", "tooltip": "The name of the input."}),
-                "value": ("STRING", {"default": "(Do Not Set Me)", "multiline": True, "tooltip": "Always leave this blank, the SwarmUI server will fill it for you."}),
+                "title": ("STRING", {"default": "My Image", "tooltip": "The name of the input."}),                
+                "image": (sorted(files), {"image_upload": True}),
                 "auto_resize": ("BOOLEAN", {"default": True, "tooltip": "If true, the image will be resized to match the current generation resolution. If false, the image will be kept at whatever size the user input it at."}),
             } | STANDARD_REQ_INPUTS,
+            "hidden":{
+                "value": ("STRING", {"default": "(Do Not Set Me)", "multiline": True, "tooltip": "Always leave this blank, the SwarmUI server will fill it for you."}),
+            },
         } | STANDARD_OTHER_INPUTS
 
     CATEGORY = "SwarmUI/inputs"
@@ -203,8 +210,11 @@ class SwarmInputImage:
     FUNCTION = "do_input"
     DESCRIPTION = "SwarmInput nodes let you define custom input controls in Swarm-Comfy Workflows. Image lets you input an image. Internally this node uses a Base64 string as input, so may not be the most friendly to use on the Comfy Workflow tab, but is very convenient to use on the Generate tab."
 
-    def do_input(self, value, **kwargs):
-        return SwarmLoadImageB64.b64_to_img_and_mask(value)
+    def do_input(self, image, value=None, **kwargs):
+        if (not value) or (value=="") or (value=="(Do Not Set Me)"):
+            return self.load_image(image)
+        else:
+            return SwarmLoadImageB64.b64_to_img_and_mask(value)
 
 
 NODE_CLASS_MAPPINGS = {
--- a/src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmInputNodes.py
+++ b/src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmInputNodes.py
@@ -1,7 +1,12 @@
 from . import SwarmLoadImageB64
 import folder_paths
-from nodes import CheckpointLoaderSimple, LoadImage
+from nodes import CheckpointLoaderSimple
 import os
+
+import node_helpers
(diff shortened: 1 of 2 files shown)


mcmonkey4eva:
This is not a terrible concept, but creates a lot of edge case issues and some misleading implications about how the node is intended to function.

OdlsCheng:
Hi @mcmonkey4eva 
I'm trying to migrate our team from ComfyUI to SwarmUI.

The Comfy Tab and Simple Tab seem suited for different users.
My expectation is: a few super users who are familiar with workflows will edit in the Comfy Tab, while most general users will use the Simple Tab to generate images.
This seems to be the intended usage pattern after SwarmUI's design evolution—and I really like this aspect.

However, for existing ComfyUI users, they’re used to executing the workflow directly, monitoring each step, and interrupting the process mid-run to make changes if the output isn't as expected.
Currently, it seems like SwarmInputImage must be use "Use This Workflow in Generate Tab" to input an image.
Does this mean we can't observe the execution process within the ComfyUI Tab?
Or did I miss something?

As someone designing workflows, I want to experiment and iterate quickly. What’s the best way to do this in SwarmUI?
Also, if I continue using this version with the Image Selector in my personal copy, what kinds of potential edge case issues might arise that I’m not aware of?

mcmonkey4eva:
`Also, if I continue using this version with the Image Selector in my personal copy, what kinds of potential edge case issues might arise that I’m not aware of?`
You prooobably wouldn't necessarily.

But... PRing something to the core of the project isn't just about your personal usage.

For example, right off the bat: This PR reordered the inputs list. That means any presaved workflows with the old version of the node will have bad values when you load it back up.
There's also a subclass to LoadImage, instead of simply calling out to it separately - that needs careful testing for side effects.

If you want to push to get this actually pulled in:
- Put value back where it was (workflow compat)
- Put 'image' on the very very bottom (workflow compat)
- Use `image=None` in the func params (api call compat)
- Clean the formatting. eg line 199 you have stray spaces, line 214 your if syntax is wonky
- Make sure you've pushed the minimum-most change possible. Is that subclassing of LoadImage strictly necessary? If not, don't do it.
- Test very very carefully

OdlsCheng:
Although the default value string is  “(Do Not Set Me)”, it doesn’t actually prevent users from modifying it manually.
If both the value field and the image selector are visible at the same time, this introduces two separate input paths, which makes unintended inputs highly likely.

Initially, I considered creating my own InputImage node in custom_nodes, but I quickly realized that in order to support input from new nodes in the Simple Tab, I would still need to modify comfy_workflow_editor_helper.js.
Since that meant I’d be changing the main repository either way, I decided at the time to modify SwarmInputImage directly instead.

Looking back, if we want to preserve compatibility with existing workflows, introducing a new SwarmInputImageWithSelector node might be the lowest-risk solution.
In my initial tests, it works fine across ComfyUI, Generate Tab, and Simple Tab.

image

Would this be considered an acceptable approach?
Aside from comfy_workflow_editor_helper.js, are there other files or components I should adjust to ensure everything integrates cleanly?

mcmonkey4eva:
`If both the value field and the image selector are visible at the same time, this introduces two separate input paths, which makes unintended inputs highly likely.`
The default value being `(Do Not Set Me)` covers this more or less well enough.
If anyone mistakenly sets it... well, there's a million other things you can mix up if you're editing raw comfy workflows, you're kinda just obligated to pay more attention on that tab.

OdlsCheng:
Before my last comment, I had already submitted a version with SwarmInputImageWithSelector.
If you would prefer to keep the changes within the original SwarmInputImage, I can adjust it.

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmInputNodes.py
```diff
@@ -186,25 +192,82 @@ def INPUT_TYPES(s):
     def do_input(self, value, **kwargs):
         return (value, )
 
+def load_image(image):
```
mcmonkey4eva: don't need the entire loadimage func here, just _call_ the standard func from inside comfy source (don't subclass just call the func)

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmInputNodes.py
```diff
@@ -186,25 +192,82 @@ def INPUT_TYPES(s):
...
             "required": {
                 "title": ("STRING", {"default": "My Image", "tooltip": "The name of the input."}),
-                "value": ("STRING", {"default": "(Do Not Set Me)", "multiline": True, "tooltip": "Always leave this blank, the SwarmUI server will fill it for you."}),
+                "value": ("STRING", {"default": s.DEFAULT_VALUE_TEXT, "multiline": True, "tooltip": "Always leave this blank, the SwarmUI server will fill it for you."}),
                 "auto_resize": ("BOOLEAN", {"default": True, "tooltip": "If true, the image will be resized to match the current generation resolution. If false, the image will be kept at whatever size the user input it at."}),
-            } | STANDARD_REQ_INPUTS,
+            } |
+            STANDARD_REQ_INPUTS |
+            {
+                "image": (sorted(files), {"image_upload": True}),
```
mcmonkey4eva, line 203: this should be under optional I think - the node is meant to work without an input to this

OdlsCheng:
> this should be under optional I think - the node is meant to work without an input to this

You're right — the `image` input should indeed be optional.

> don't need the entire loadimage func here, just _call_ the standard func from inside comfy source (don't subclass just call the func)

I have initially looked into it, but couldn't find a suitable static method for loading images.
The `load_image(self, image)` function in `LoadImage` is an instance method.

Currently, `self` isn’t used in that method, so calling it with `LoadImage.load_image(None, image)` might work.
However, since it's an instance method, passing `None` as `self` feels unsafe — there's no guarantee that `LoadImage` won't start using `self` internally in the future.

While duplicating `load_image` is admittedly a bit ugly, I concluded it's the safer option: it avoids breaking if `LoadImage.load_image()` is changed in future ComfyUI updates.

mcmonkey4eva:
You just `LoadImage().load_image(...)`

OdlsCheng:
> You just LoadImage().load_image(...)

Cool! I’m a C# developer, so I never considered using object instances in this one-off way.
If that’s a conventional and reasonable approach in Python, then it actually looks like a good solution.

> this should be under optional I think - the node is meant to work without an input to this

After making `image` optional, I noticed that the UI no longer displays the image upload field.
After some digging, I finally tracked down the source of that behavior—it’s not even in the ComfyUI repository, but rather in ComfyUI_frontend

In [uploadImage.ts](https://github.com/Comfy-Org/ComfyUI_frontend/blob/4b75528c39351abd1e14bcce9d7708b45acbb352/src/extensions/core/uploadImage.ts):
```ts
const found = Object.entries(required).find(([_, input]) => isMediaUploadComboInput(input))
if (found) {
  const [inputName, inputSpec] = found
  required.upload = createUploadInput(inputName, inputSpec)
}
```
The frontend only looks for `"image_upload"` inputs inside `Object.entries(required)` to show the `UploadInput`.

To work around this, I might need to implement a corresponding JS extension for `SwarmInputImage`.
I'm less familiar with JS than with Python, so it may take me some time to explore this.
If anyone has suggestions or guidance, I would greatly appreciate it!

mcmonkey4eva:
That's... unfortunate, but okay in that case leave it in required and we'll just have to make sure that's handled

PR #903, opened 2025-07-18
bump nuget versions

kalebbroo:
- Update nugets to latest versions
- Rebuilt and checked for errors or outdated logic


mcmonkey4eva:
what

PR #904, opened 2025-07-19
do not remove 2x items from chunkSize from remainingFiles when building item list

bman654:
I had 646 images in my list and noticed the last 33 or so were not being shown.  Caused by a bug where 2x chunkSize were being removed from list instead of 1x chunkSize (splice removes the first chunkSize, then slice removed another chunkSize)


mcmonkey4eva:
Good catch! `splice` modifying the original array and `slice` not doing that must've gotten mixed up

PR #908, opened 2025-07-20
Item browser performance improvements

bman654:
ok I was getting frustrated by all the flashing and slow updating of the Image History tab when refreshing it.  It is ESPECIALLY bad if I am in a nested folder.  If I am in a/b/c/d, then refreshing causes it to flash with the contents of a, then a/b, then a/b/c and finally a/b/c/d.  And if I have 1000 images those flashes are slow.

So here are some performance updates.  I put each specific update in a separate commit in case you have any reservations we can just cherry pick the ones you will take.  They are generally in order from least-risky to most-risky (only the last one is actually risky).

If you take them all, for my test case (nested folder with 646 images), clicking Refresh originally took 11+ seconds.  Now it takes only 2.2 seconds. 

Here are a bunch of profiler screenshots....

### Original Behavior

If I am viewing this nested folder with 646 images:

image

And I click Refresh, the original code with all its flashing ends up fetching the ImageList from the server 4 times and  takes **11+ seconds** to settle down.  This is what the flamegraph looks like:

image

Lots of Reflow.  Lots of script cpu time.  The image list is re-rendered 4 times I think.

### Updated Behavior

With all the changes, the refresh runs in 2.2 seconds and the profile looks like:

image

We've removed many of the extra calls to ListImages.  We've removed most of the reflow.  We've reduced the script-time spent generating html.  We've reduced the number of times the list gets re-rendered.

# Progressive improvements

if you zoom in on a single render of the root folder with 1200 images, it takes about **2.4s**.

image

There is a huge chunk when the html is generated, and then a bunch of trailing reflows as images are loaded.

## Commit 1 - thumbnail aspect ratio

All of the trailing reflow after render was caused by the images loading and the thumbnail aspect ratio being modified from 1:1 to whatever the image aspect ratio was.  But for Image History, we usually know what the ratio is for each image when we are initializing the thumbnail.  So just setting that and using some css tricks to update it without JavaScript when the image finally arrives (if it does not match the ratio we pre-computed) makes it so the thumbnails end up being correctly sized from the start and all the reflow on image load vanishes.  It makes it faster and as a bonus the user doesnt see the tiles moving around anymore.

This drops the single-render time to roughly **900ms**.

image

## Commit 2

I was trying to eliminate the reflow caused by setting scrollTop.  I changed the code to only set it if it is > 0.  All this really did was move the reflow to `makeVisible`, but that's ok....

## Commit 3

Because now that makeVisible is the bottleneck, we can optimize it.  Instead of a tight loop of (read DOM, modify DOM) which causes a tight loop of reflow, we instead do all the DOM reading first, then do all the DOM modifications second.  I've lost my profile, but I think this got the render time down to around 750ms.

## Commit 4

I notice a lot of time is spent parsing html. It's the code that converts an html description into a text title.  html->text is pretty slow if using the DOM to do it.  And why do it on 1000 items just in case someone is going to move their mouse over 1 or 2 and get a tooltip?  So I modified the code to do it on demand only when the user actually moves the mouse into the element.

This change got the single-render time down to roughly 600ms.

## Commit 5

This one is the most risky.  I tackled the problem of the multiple calls to GetItemList and multiple renders of the items.  Now it makes one call to get the root files/folders (used to populate the folder list) and then only makes calls when it gets to the depth limit and hasnt reached the target folder yet.  And while it is traversing the tree, it refrains from rendering the items.  Only once it hits the target tree node does it make a final call to get the files at this depth, then renders them.

This commit is only risky in the sense that it's possible I broke the refresh on some tab I'm not familiar with.  Hopefully you can test it as well.  I've tried it with a couple of different depth settings (1, 2, 6) and it seems to work correctly.

### DIFF
--- a/src/wwwroot/css/genpage.css
+++ b/src/wwwroot/css/genpage.css
@@ -252,8 +252,6 @@ body {
 .image-block img, .img-block video {
     display: block;
     margin: auto;
-    padding-left: 0.4rem;
-    padding-right: 0.4rem;
 }
 .image-block-placeholder img {
     filter: blur(5px);
@@ -264,6 +262,7 @@ body {
 .image-block-legacy {
     position: relative;
     height: 11.5rem;
+    width: 10rem;
     padding-top: 0.1rem;
     padding-left: 0.1rem;
     padding-right: 0.1rem;
@@ -273,12 +272,15 @@ body {
 }
 .image-block-small {
     height: 7rem;
+    width: 7rem;
 }
 .image-block-big {
     height: 17.5rem;
+    width: 16rem;
 }
 .image-block-giant {
     height: 27.5rem;
+    width: 26rem;
 }
 .image-block-legacy img {
     height: 8rem;
--- a/src/wwwroot/css/genpage.css
+++ b/src/wwwroot/css/genpage.css
@@ -364,7 +364,6 @@ body {
 }
 .image-preview-text {
     font-family: var(--font-monospace);
-    /* this dark magic will make the text grow as big as the parent container is and then wrap (instead of growing the container even more) */
     min-width: 100%;
     max-width: 0;
 }
--- a/src/wwwroot/css/genpage.css
+++ b/src/wwwroot/css/genpage.css
@@ -252,6 +252,8 @@ body {
 .image-block img, .img-block video {
     display: block;
     margin: auto;
+    padding-left: 0.4rem;
+    padding-right: 0.4rem;
 }
 .image-block-placeholder img {
     filter: blur(5px);
@@ -262,7 +264,6 @@ body {
 .image-block-legacy {
     position: relative;
     height: 11.5rem;
-    width: 10rem;
     padding-top: 0.1rem;
     padding-left: 0.1rem;
     padding-right: 0.1rem;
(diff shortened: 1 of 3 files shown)


brendanhoar:
I'm excited!

But also, I don't really count here, lol.

mcmonkey4eva:
I'm having a hard time validating things, as
(A) on my machine it's basically no issue even with ~5000 images/videos in a history tab lol. Not sure if it's a browser choice thing (Firefox is pretty great, Chrome is very slow on some simpler update funcs I've found, especially so when debugging is active) or just my hardware is too strong.
(B) in both your PR as originally submitted, and with my modifications to it, there's occasional visible reflows while loading (ie I'll watch it jump from a late resize) (probably because i got weird ass experimental nonsense filling my image history tbh)

-----------

I've removed everything that's, *not proper PRing*, notably stray comments (should be commit comments not code comments), stray css changes (???), and stray code changes (do things the way the existing code does unless you have a reason to be changing how it's done, otherwise you may be causing accidental regressions on intentional decisions, or creating consistency-related issues). The hijack on `update` is... maybe if you were trying to mod it in entirely externally an approach like that makes sense? But for core code that's wild. I replaced it with a simple boolean and some `rerender()`s.

Ftr btw the refresh path building code that's currently there in the core is a very silly hack already, triggering clicks through the tree to load into it. Optimizing it is almost redundant vs just replacing it with a thing that loads straight to the target path in one go.

---------

That all said... neat work! Always love some optimization. And the aspect ratio precalc has immediate visual improvements for compatible images, which is pretty cool.

PR #929, opened 2025-07-30
Including a missing dependency (Pip)

monksy:
Added pip otherwise the UI complains it isn't included in the docker container.


mcmonkey4eva:
Does it really complain? I'd expect `python3.11-dev` to include pip

monksy:
It does on the start. I don't have a screenshot.

mcmonkey4eva:
yeah it seems it does actually need that - that's... really odd.

monksy:
I frequently have that issue everytime I install python as well.. oh well.

PR #931, opened 2025-07-30
Support Numbers Larger Than int32 for `T2IParamDataType.INTEGER`

Odls:
The `T2IParamDataType.INTEGER` parameter corresponds to Python’s integer type, not C#’s `int32`.
To accommodate larger numbers when passing values in C#, some parts of the code already treat it as a `long`.

For example:
In `T2IParamSet.cs`:
```csharp
object obj = param.Type switch
{
    T2IParamDataType.INTEGER => param.SharpType == typeof(long) ? long.Parse(val) : int.Parse(val),
```

In `T2IParamType.cs`:
```csharp
public static Type DataTypeToSharpType(T2IParamDataType t)
{
    return t switch {
        T2IParamDataType.INTEGER => typeof(long),
```

However, in both `ComfyUIBackendExtension.cs` and `ComfyUIAPIAbstractBackend.cs`, the received parameters were being cast to `int32`, causing parse failures when users input large numbers.
A common scenario is when a user wants to reuse the seed from an existing image, which can often be a 14–16 digit number—too large to fit in `int32`, and therefore not accepted by `SwarmInputInteger`.

This commit updates all relevant usages of `T2IParamDataType.INTEGER` to use `long`, and verifies that `SwarmInputInteger` can now successfully pass a 16 digit number seed.

### DIFF
--- a/src/BuiltinExtensions/ComfyUIBackend/ComfyUIAPIAbstractBackend.cs
+++ b/src/BuiltinExtensions/ComfyUIBackend/ComfyUIAPIAbstractBackend.cs
@@ -739,8 +739,8 @@ public abstract class ComfyUIAPIAbstractBackend : AbstractT2IBackend
                     }
                     if (type.Type == T2IParamDataType.INTEGER && type.ViewType == ParamViewType.SEED && long.Parse(val.ToString()) == -1)
                     {
-                        long max = (long)type.Max;
-                        return $"{Random.Shared.NextInt64(0, max <= 0 ? long.MaxValue : max)}";
+                        int max = (int)type.Max;
+                        return $"{Random.Shared.Next(0, max <= 0 ? int.MaxValue : max)}";
                     }
                     if (val is T2IModel model)
                     {
--- a/src/BuiltinExtensions/ComfyUIBackend/ComfyUIAPIAbstractBackend.cs
+++ b/src/BuiltinExtensions/ComfyUIBackend/ComfyUIAPIAbstractBackend.cs
@@ -739,8 +739,8 @@ public abstract class ComfyUIAPIAbstractBackend : AbstractT2IBackend
                     }
                     if (type.Type == T2IParamDataType.INTEGER && type.ViewType == ParamViewType.SEED && long.Parse(val.ToString()) == -1)
                     {
-                        int max = (int)type.Max;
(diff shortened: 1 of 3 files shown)


src/BuiltinExtensions/ComfyUIBackend/ComfyUIAPIAbstractBackend.cs
```diff
@@ -739,8 +739,8 @@ string fillDynamic()
                     }
                     if (type.Type == T2IParamDataType.INTEGER && type.ViewType == ParamViewType.SEED && long.Parse(val.ToString()) == -1)
                     {
-                        int max = (int)type.Max;
-                        return $"{Random.Shared.Next(0, max <= 0 ? int.MaxValue : max)}";
+                        long max = (long)type.Max;
+                        return $"{Random.Shared.NextInt64(0, max <= 0 ? long.MaxValue : max)}";
```
mcmonkey4eva: this very intentionally used int.MaxValue here and should be put back

src/BuiltinExtensions/ComfyUIBackend/ComfyUIBackendExtension.cs
```diff
@@ -197,11 +197,11 @@ public T2IParamType DynamicParamGenerator(string name, T2IParamInput context)
...
                         string seedClean(string prior, string newVal)
                         {
-                            int parsed = int.Parse(newVal);
+                            long parsed = long.Parse(newVal);
                             if (parsed == -1)
                             {
-                                int max = (int)type.Max;
-                                parsed = Random.Shared.Next(0, max <= 0 ? int.MaxValue : max);
+                                long max = (long)type.Max;
+                                parsed = Random.Shared.NextInt64(0, max <= 0 ? long.MaxValue : max);
```
mcmonkey4eva: (same as above for both in this file)

src/BuiltinExtensions/ComfyUIBackend/ComfyUIBackendExtension.cs
```diff
@@ -417,7 +417,7 @@ public static void OnSwarmBackendAdded(SwarmSwarmBackend backend)
     }
 
     public static LockObject ValueAssignmentLocker = new();
-
```
mcmonkey4eva: minor but please avoid unrelated line edits

PR #933, opened 2025-07-30
Add finalization handlers for prompt preprocessing steps

bman654:
This is not needed by the Core application but acts as an extension point for Extensions that want to manipulate the input params after all of the wildcard, random, and other prompt-modification tags have been processed.

I was previously hooking into the pipeline in a wonky way (via the `PreGenerateEvent` event) but that event only fired during generation and there was no equivalent way to hook into the process when clicking the "Generate Workflow from Generator tab" button on Comfy tab.   Also some of the processing I wanted to run was async and I felt bad everytime I wrote `Task.Result` to block until the task completed.


mcmonkey4eva:
Valid idea but I don't love the details - replaced with `LateSpecialParameterHandlers`

re async: yeah you're doing async things while a gen is prepping to run, so, you are potentially delaying the user request, so, you should feel weird and have to do some weird taskhacking about it.

Issue #961, opened 2025-08-16
Qwen-image returns error when used as tiled refiner


willhsmit:
### Expected Behavior

When used as a refiner in Refine / Upscale with 'Refiner Control percentage' > 0 and 'Refiner Do Tiling' checked, Qwen should work and refine the image.

### Actual Behavior

Swarm throws an exception 

ComfyUI execution error: too many values to unpack (expected 4)

Basically, it looks like split_latent_tensor and stitch_latent_tensors are expecting the passed-in latents to have 4 dimensions, but they have 5.

From a printf, we see the latent tensor passed into split has this shape 
Hidream: torch.Size([1, 16, 278, 216])
Qwen image: torch.Size([1, 16, 1, 278, 216])

Full stack trace is in the debug logs section.

### Steps to Reproduce

Use an init image.

Following parameters:
Init Image Creativity: 0, 
Refiner Control Percentage: 0.2, 
Refiner Method: Post-Apply (Normal), 
Refiner Upscale: 2, 
Refiner Upscale Method: Model: 4xNomos8kDAT.pth, 
Refiner Do Tiling: true

Model: qwen-image-Q3_K_M.gguf
LoRAs: Qwen-Image-Lightning-8steps-V1.1-bf16 : 1

We get an error on the tiling step.

### Debug Logs

10:37:30.390 [Warning] [ComfyUI-0/STDERR] Traceback (most recent call last):
10:37:30.391 [Warning] [ComfyUI-0/STDERR]   File "C:\Users\willh\SwarmUI\dlbackend\comfy\ComfyUI\execution.py", line 496, in execute
10:37:30.391 [Warning] [ComfyUI-0/STDERR]     output_data, output_ui, has_subgraph, has_pending_tasks = await get_output_data(prompt_id, unique_id, obj, input_data_all, execution_block_cb=execution_block_cb, pre_execute_cb=pre_execute_cb, hidden_inputs=hidden_inputs)
10:37:30.392 [Warning] [ComfyUI-0/STDERR]                                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #962, opened 2025-08-16
Handle latents of dimension > 4 in tiled refining

willhsmit:
See https://github.com/mcmonkeyprojects/SwarmUI/issues/961.

When using Qwen-image as a tiled refiner, it appears to pass latents of dimension > 4 into split_latent_tensor and stitch_latent_tensors. Swarm fails at the splitting/stitching step of doing tiled refinement.

Writing a debug line, Hidream gives split_latent_tensor: torch.Size([1, 16, 278, 216])
Qwen gives split_latent_tensor: torch.Size([1, 16, 1, 278, 216])

This PR handles latents of dimension > 4, without making any assumption about what the new dimensions mean except that the height and width are the last two dimensions.


willhsmit:
Tested with input image being 2x-ed:

Original :
Friederike_von_Preussen_-_1796

Hidream: (4 dimensions, worked before, still works)
1026--hidream-i1-dev-Q4_K_S-186633719

Qwen-image: (5 dimesnions, didn't work before, now works)
1022--qwen-image-Q3_K_M-659782128

Random remark:
It's surprising how much modification Qwen does at Refiner control percentage 0.2.

PR #963, opened 2025-08-17
feat: make trigger phrases clickable for easy copying

x0x0b:
### Problem
Model trigger phrases are currently displayed as plain text, requiring users to manually copy them to use in prompts.

### Solution
Add interactive trigger phrase functionality:

- **Click-to-copy**: Trigger phrases are now clickable with visual feedback
- **Keyboard accessibility**: Support Enter/Space keys for screen readers and keyboard navigation
- **User feedback**: Success/error notifications confirm copy operations
- **Multi-phrase support**: Handle comma-separated phrases (`,, ` delimiter) individually

### Example
`test1, test2, test3,, testA, testB, testC,, <b>sanitized</b>,, <script>alert('1')</script>`

ss2

### Ask
Since I am not very experienced with frontend implementation, please pay special attention to whether the XSS protection is sufficient. In particular, I would appreciate it if you could confirm that output escaping are properly implemented.
In the parts I newly implemented, I am using `escapeHtml` instead of `safeHtmlOnly`.

### DIFF
--- a/src/wwwroot/js/genpage/gentab/models.js
+++ b/src/wwwroot/js/genpage/gentab/models.js
@@ -468,22 +468,17 @@ class ModelBrowserWrapper {
 
     createCopyableTriggerPhrase(phrase) {
         let copyPhrase = phrase;
-        if (getUserSetting('ui.copytriggerphrasewithtrailingcomma', true) && !phrase.endsWith(',')) {
-          copyPhrase += ',';
+        if (getUserSetting('ui.copytriggerphrasewithtrailingcomma', false) && !phrase.endsWith(',')) {
+          copyPhrase += ', ';
         }
-        const safePhrase = escapeHtmlNoBr(escapeJsString(phrase));
-        const safeCopyPhrase = escapeHtmlNoBr(escapeJsString(copyPhrase));
+        let safePhrase = escapeHtmlNoBr(escapeJsString(phrase));
+        let safeCopyPhrase = escapeHtmlNoBr(escapeJsString(copyPhrase));
         return `${safePhrase}<button title="Click to copy" class="basic-button trigger-phrase-copy-button" onclick="copyText('${safeCopyPhrase}');doNoticePopover('Copied!', 'notice-pop-green');">&#x29C9;</button>`;
     }
 
     formatTriggerPhrases(val) {
-        if (val.includes(';')) {
-            const phrases = val.split(';').map(phrase => phrase.trim()).filter(phrase => phrase.length > 0);
-            return phrases.map(phrase => this.createCopyableTriggerPhrase(phrase)).join('');
-        }
-        else {
-            return this.createCopyableTriggerPhrase(val);
-        }
+        let phrases = val.split(';').map(phrase => phrase.trim()).filter(phrase => phrase.length > 0);
+        return phrases.map(phrase => this.createCopyableTriggerPhrase(phrase)).join('');
     }
 
     describeModel(model) {
@@ -585,9 +580,7 @@ class ModelBrowserWrapper {
         let searchableAdded = '';
         if (model.data.is_supported_model_format) {
             let getLine = (label, val) => {
-                const content = label === 'Trigger Phrase'
-                    ? `${this.formatTriggerPhrases(val)}`
-                    : `${val == null ? '(Unset)' : safeHtmlOnly(val)}`;
+                let content = val == null ? '(Unset)' : (label == 'Trigger Phrase' ? this.formatTriggerPhrases(val) : safeHtmlOnly(val));
                 return `<b>${label}:</b> <span>${content}</span><br>`;
             };
             let getOptLine = (label, val) => val ? getLine(label, val) : '';
--- a/src/wwwroot/css/genpage.css
+++ b/src/wwwroot/css/genpage.css
@@ -1632,5 +1632,7 @@ body {
 .trigger-phrase-copy-button {
     display: inline-flex;
     align-items: center;
-    height: 1em;
+    font-size: 90%;
+    height: 1.6em;
+    padding-top: 0 !important;
 }


mcmonkey4eva:
That `,,` syntax feels a bit arbitrary?

And: there are existing 'click to copy' buttons, you should try to reuse both UI design and code from those
image
both for user-facing consistency, and for code maintainability (if there's two different sections of code that do the same thing, if there's a bug, it might get fixed in one but not the other or something like that)

x0x0b:
Thanks for the review!

I realized that my implementation with the `,,` delimiter was based on a misunderstanding — it’s not something that applies to all models.
(In the model I downloaded from Civitai, the `trainedWords` field was set like `["trigger1,", "trigger2,"]`, which led me to believe that the trigger phrase was being treated as `"trigger1,, trigger2,"`. That confused me and I mistakenly generalized it.)

That said, I think the current implementation here could still be problematic:
https://github.com/mcmonkeyprojects/SwarmUI/blob/46081c9ced30f52b78a2062fb84b08ecabf800e4/src/wwwroot/js/genpage/utiltab.js#L309

Since it relies on splitting by a plain comma, it cannot reliably distinguish between a delimiter and commas that are actually part of a trainedWords string.
For example, this model on Civitai (https://civitai.com/models/1866531/rooftop-pool) has `"trainedWords": ["r00fp00l", "outdoors, ..."]`, but after download it becomes a single string `"r00fp00l, outdoors, ..."`, which makes the comma ambiguous.
Wouldn’t it be safer to use a different separator (for example `;` or `|`)? That would also make the change in this PR easier to support.

Regarding your note about reusing the existing copy-to-clipboard button and logic — I fully agree. I will refactor this part of the implementation to align with the project’s standard component and avoid duplicating functionality.

mcmonkey4eva:
`;` as an intentional separator seems more appropriate, I'd be good with changing the metadata autopopulator to produce that. There's a few different places it'd need to be handled (notably where you linked, and the metadata loading in t2imodelhandler, and... probably documented somewhere), and some special considerations (notably: `<trigger>` probably shouldn't emit stray semicolons to the actual prompt. Maybe replace back to `,`? Or only pick first? Or... idk. Probably replace back to `,`).

For the record: generally models are expected to have "a trigger phrase", and lists of many different ones are kinda weird (and very often when they exist they are a simple mistake on the part of a civitai uploader), but valid enough to exist in some cases.
In the rooftop pool model you linked, "r00fp00l" is a deeply cursed trigger word, and the other part is just some suggested chunk of prompt text to encourage the model to figure out what the hell you're talking about.

x0x0b:
Thank you for the detailed explanation.

There turned out to be more considerations than I initially expected when modifying Trigger Phrases during download, and the changes became rather extensive.
Also, although it might not have a major impact, the fact that an arbitrary `;` would be inserted—making it different from the Trigger Phrases in models that have already been downloaded—feels a bit undesirable.

For now, I feel it might be better not to change the save logic for Trigger Phrases, and instead implement it with a button that simply copies the entire Trigger Phrase. What do you think?

mcmonkey4eva:
That'd be good too yeah

mrblomblo:
I'm not sure if my input is of use, but I think a lot of people (me included) would appreciate having the ability to copy different trigger phrases, not just copy all trigger phrases as one. This is especially useful for character pack LoRas or character LoRas where multiple different outfits or characters are available.  
For example: https://civitai.com/models/289093/that-time-i-got-reincarnated-as-a-slime-sdxl-version-upgraded  
Where trigger phrase 1 is "Albis", trigger phrase 2 is "Alice", etc.

I do, however, understand from the previous comments that it would be difficult to implement in a good way.

x0x0b:
I implemented this in commit 29c36a778605c65b1fd83d4c79d54bd21fc54959, using the existing copy button implementation here as a reference: https://github.com/mcmonkeyprojects/SwarmUI/blob/eab5dfdcd255d01d8e755b75f8c4c5baadeed45a/src/wwwroot/js/genpage/helpers/metadatahelpers.js#L188

Could you take a look and let me know what you think? For cases with multiple trigger phrases, I feel the previous appearance looked better.

before:
スクリーンショット 2025-08-21 7 23 44

after:
スクリーンショット 2025-08-21 7 14 08

P.S. Sorry, some of my commits are showing as "Unverified".

mrblomblo:
Hmm, If we're going with the current version (with the copy buttons separate) I think, if possible, the commas should appear after the copy button, not before.  
Before:  
image  
After:  
image

Though, I personally think something like this would look better and be clearer to the end user of what they're copying (and that it's a copy action):  
image
Where you add the trigger phrase text to the copy button. I have just used the inspect element tool to make the mockups, but I would guess replacing:
```js
return `<span>${safePhrase}</span><button title="Click to copy" class="basic-button" onclick="copyText('${safePhrase}');doNoticePopover('Copied!', 'notice-pop-green');">&#x29C9;</button>`;
```  
with something like the following:
```js
return `<button title="Click to copy" class="basic-button" onclick="copyText('${safePhrase}');doNoticePopover('Copied!', 'notice-pop-green');">${safePhrase} &#x29C9;</button>`;
```  
could work? (at line 499 in `src/wwwroot/js/genpage/gentab/models.js`)

Edit: Tested it by editing my `models.js`, and it works, but the commas are still in the copied trigger phrases and UI, not sure if that's desireable or not. Could get rid of them by deleting line 497
```js
const phraseWithComma = phrase.endsWith(',') ? phrase + ' ' : phrase + ', ';
```
and replacing "phraseWithComma" with "phrase" on what was previously line 498
```js
// old
const safePhrase = escapeHtmlNoBr(escapeJsString(phraseWithComma));
// new
const safePhrase = escapeHtmlNoBr(escapeJsString(phrase));
```

Doing that would get you the following (comma on last one because it is there in the trigger phrase):
image

x0x0b:
Thank you! I really like it.

The reason I add a trailing `, ` is to make it smoother to continue typing the next prompt after pasting the phrase, but is that a bit too presumptuous? Having the comma also makes the copy button look worse.
I considered keeping the comma out of the visible text and only appending it at copy/paste time, but that behavior isn't very intuitive, so it might be better to remove it.

mrblomblo:
Great!

Yeah I agree that having the trailing comma makes it easier, but like you said; it makes the copy button not as visually appealing, and ideally the text you see should be what you copy (at least I think that makes sense).  
The best-case scenario would IMO be to add a toggle-able user setting for appending a ", " to the copied text while not displaying it in the button. Though, that would add unnecessary complexity.

I think this should probably be up to mcmonkey to decide though, but in my opinion, having no trailing comma is better :)

x0x0b:
Just a quick note: for example, if a Trigger Phrase is set like `test1; test2,; test3, ; test4`, you may see commas appear inconsistently (as shown in the attached screenshot). 
I believe this happens because whether a Trigger Phrase ends with a comma is not standardized on Civitai, so the presence or absence of a comma can vary between models.

スクリーンショット 2025-08-21 8 28 22

mrblomblo:
Perhaps you could modify the "phrasesWithCommas" that was deleted, so it removes trailing commas instead of adding them? I'm not good at JS so my example may not be good, but for me it works with trailing commas and trailing commas + spaces:
```js
const phraseWithoutComma = phrase.endsWith(',') ? phrase.slice(0, -1) : phrase;
const safePhrase = escapeHtmlNoBr(escapeJsString(phraseWithoutComma));
```

But now that I think about it, the user still sees the comma in the trigger phrase when editing the metadata, so IDK if this is the best approach, since some users may want to have the commas?

x0x0b:
I've come to feel that both adding or removing trailing commas would fail to respect the Trigger Phrase in the metadata, so it's better not to do that (users can edit the metadata themselves). 
If such behavior is offered, I personally think it would be better to provide it as an optional, configurable setting.

x0x0b:
I added an option to append a trailing comma.
I wasn't sure whether to put it in the UI section or the metadata section, so I added it to the UI section.
Is it okay to add it here? 744de8befc1811796ee249de97d07c983eb12b4c

mrblomblo:
> The best-case scenario would IMO be to add a toggle-able user setting for appending a ", " to the copied text while not displaying it in the button. **Though, that would add unnecessary complexity.** (https://github.com/mcmonkeyprojects/SwarmUI/pull/963#issuecomment-3208350730)

My bad, I didn't know it was that easy to add to the settings lol  
Looks great! If it were me, I'd also have added it to the UI section :)

Regarding the trailing commas, maybe it is possible to strip them from the metadata directly when imported from CivitAI? Not sure if that's a good idea or not, but it'd be a solid feature IMO (though for another PR of course).

src/Core/Settings.cs
```diff
@@ -368,6 +368,9 @@ public class UserUIData : AutoConfiguration
 
             [ConfigComment("Comma-separated list of fields to display in the preset Details view.\nUse 'name' for the preset name, 'path' for the full preset path, 'description' for the description, or 'params' for the param list.\nIf unset, will act as 'path,description,params'")]
             public string PresetListDetailsFields = "";
+
+            [ConfigComment("If enabled, trigger phrases are copied with a trailing comma added.\nIf disabled, trigger phrases are copied as-is without any trailing comma.\nThis is useful when copying them to prompts.")]
+            public bool CopyTriggerPhraseWithTrailingComma = true;
```
mcmonkey4eva: this should default off, it's slightly weird behavior to add a comma that's not there in the original data

src/wwwroot/js/genpage/gentab/models.js
```diff
@@ -466,6 +466,38 @@ class ModelBrowserWrapper {
...
     }
 
+    createCopyableTriggerPhrase(phrase) {
+        let copyPhrase = phrase;
+        if (getUserSetting('ui.copytriggerphrasewithtrailingcomma', true) && !phrase.endsWith(',')) {
+          copyPhrase += ',';
+        }
+        const safePhrase = escapeHtmlNoBr(escapeJsString(phrase));
+        const safeCopyPhrase = escapeHtmlNoBr(escapeJsString(copyPhrase));
+        return `<button title="Click to copy" class="basic-button" onclick="copyText('${safeCopyPhrase}');doNoticePopover('Copied!', 'notice-pop-green');">${safePhrase} &#x29C9;</button>`;
```
mcmonkey4eva: the text should not be inside a button

src/wwwroot/js/genpage/gentab/models.js
```diff
@@ -466,6 +466,38 @@ class ModelBrowserWrapper {
...
+    }
+
+    formatTriggerPhrases(val) {
+        if (val.includes(';')) {
+            const phrases = val.split(';').map(phrase => phrase.trim()).filter(phrase => phrase.length > 0);
+            return phrases.map(phrase => this.createCopyableTriggerPhrase(phrase)).join('');
+        } else {
```
mcmonkey4eva: please match code format to existing code

src/wwwroot/js/genpage/gentab/models.js
```diff
@@ -466,6 +466,38 @@ class ModelBrowserWrapper {
...
+            return this.createCopyableTriggerPhrase(val);
+        }
+    }
+
+    createInfoLine(label, value) {
+        if (value == null) {
+            return `<b>${escapeHtml(label)}:</b> <span>(Unset)</span><br>`;
```
mcmonkey4eva: this code has 3 copies of the same thing, and missing the span on one of them.

src/wwwroot/js/genpage/gentab/models.js
```diff
@@ -564,7 +596,9 @@ class ModelBrowserWrapper {
         }
         let searchableAdded = '';
         if (model.data.is_supported_model_format) {
-            let getLine = (label, val) => `<b>${label}:</b> <span>${val == null ? "(Unset)" : safeHtmlOnly(val)}</span><br>`;
+            let getLine = (label, val) => {
+                return this.createInfoLine(label, val);
```
mcmonkey4eva: this makes no sense as written

src/wwwroot/js/genpage/gentab/models.js
mrblomblo: I don't want to come off as rude or argumentative, I'm just genuinely curious as to why it shouldn't be inside the button. Having it in the button makes the whole trigger phrase clickable without having to write unnecessary code that would accomplish the same thing as the current suggested implementation.

I understand this isn't the only alternate way of doing it, but it just looks plain bad (along with being worse for accessibility) with the button next to the trigger phrase compared to having it inside the button. Imagine you have a really long trigger phrase (for example a specific outfit), you'd need to scroll down a bit to be able to click the "Copy" button. With the current suggested implementation, you'd just need to click the trigger phrase with no need for scrolling.
image

x0x0b: I personally agree with mrblomblo-san. As shown in the attached screenshot, Civitai’s UI also places the trigger phrase text inside the button, so it doesn’t feel out of place.
スクリーンショット 2025-08-23 7 06 23

Of course, if there are semantic or implementation concerns (for example, side effects from putting text inside a `<button>`), I’m happy to consider alternative approaches that take those issues into account. I hope this is helpful.

src/Core/Settings.cs
x0x0b: Thank you for reviewing!

I fixed it in 2e1e5e7.

src/wwwroot/js/genpage/gentab/models.js
x0x0b: I fixed it in f55134f.

src/wwwroot/js/genpage/gentab/models.js
x0x0b: I fixed it in 7dc8aa0. 🙇🏻

src/wwwroot/js/genpage/gentab/models.js
x0x0b: I moved `createInfoLine` function into it. 9265526

src/wwwroot/js/genpage/gentab/models.js
x0x0b: In the original implementation the label wasn't using `escapeHTML()` as in `<b>${label}:</b>`. Should I revert it?
https://github.com/x0x0b/SwarmUI/blob/a50ab92c1365d06229bee9b7a2186f2cd22c027f/src/wwwroot/js/genpage/gentab/models.js#L587

mcmonkey4eva: that's not needed, labels are hard defined in code

x0x0b: Thanks. I removed it. 🙇🏻  9ed8235

src/wwwroot/js/genpage/gentab/models.js
mcmonkey4eva: Text needs to not be in buttons.

The buttons can probably be made nicer. At least made to not be taller than everything else is lol.

mrblomblo: Buttons in general (including other elements styled as buttons) or just `<button>` elements? I'm honestly just grasping at straws here. Having the text outside the button you clicked will likely confuse users about if it will copy the text before or after it (especially if there is no real visual indicator apart from a single comma).

You could of course style it so the button and text are in the same button-like "container", but still separate, which would make it look better (only copy button clickable, text not clickable). But this *could* lead to confusion from people attempting to click the *text* to copy. (shite example pic, but you get the point)
image

x0x0b: I fixed it in d9c19c6. If you have a better CSS implementation, please let me know. 🙇🏻 
スクリーンショット 2025-08-23 8 29 17

mrblomblo: _[Redacted because I now understand what monkey meant by "text in buttons" after discussion in the Discord server]_

PR #966, opened 2025-08-19
Small UI Improvements

mrblomblo:
* Card body no longer clips the card border (rounded bottom card body corners)  
  image

* Image info (image extras) no longer has 0 margin to the right  
  image  
  The view with the image next to the image info (image extras) should not be affected, at least it wasn't in my testing.  
* Image extras param blocks can no longer overflow text, instead clipping it and displaying ellipses  
  image  
  In my testing it only has an effect if you are in the view mode where the image info (image extras) and the image are next to each other, as it seems like it already properly cuts off the text when in the view mode where the image is above the image info.  
* Model block images (Small Card, Card, Big Card) are now centered  
  Small Cards:  
  image  
  Regular Cards:  
  image  
  Big Cards:  
  image

### DIFF
--- a/src/wwwroot/css/genpage.css
+++ b/src/wwwroot/css/genpage.css
@@ -220,6 +220,7 @@ body {
     min-width: 20rem;
     vertical-align: top;
     margin-left: 0.5rem;
+    margin-right: 0.5rem;
     max-height: 10rem;
 }
 .extras-wrapper-sideblock .current-image-data {
@@ -391,6 +392,7 @@ body {
 .model-block img {
     max-height: 9rem;
     width: 9rem;
+    height: 9rem;
     display: inline-block;
     margin: 0.5rem;
     margin-bottom: 0px;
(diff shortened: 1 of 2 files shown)


FurkanGozukara:
@mrblomblo nice. can you also add batch delete of presets? perhaps with checkboxes and check all option

mrblomblo:
> @mrblomblo nice. can you also add batch delete of presets? perhaps with checkboxes and check all option

Hmm, that would need to be in a separate PR I think. But I am not experienced enough with programming that I feel like it would be appropriate for me to make that feature :/

Edit: You could create a feature request, though!

FurkanGozukara:
> > @mrblomblo nice. can you also add batch delete of presets? perhaps with checkboxes and check all option
> 
> Hmm, that would need to be in a separate PR I think. But I am not experienced enough with programming that I feel like it would be appropriate for me to make that feature :/
> 
> Edit: You could create a feature request, though!

ok i am trying to implement myself lets see if i can do

mrblomblo:
> ok i am trying to implement myself lets see if i can do

Cool! Good luck :)

PR #967, opened 2025-08-19
Batch preset delete feature implemented

FurkanGozukara:
Implemented batch delete button

image

When you click the button it will be like this

Automatically all checked

image

Then uncheck the ones you don't want get deleted and then click batch delete again

It will ask again

image

Click ok and all deleted

### DIFF
--- a/src/wwwroot/css/genpage.css
+++ b/src/wwwroot/css/genpage.css
@@ -609,6 +609,29 @@ body {
     border: 1px solid var(--box-selected-border);
     background-color: var(--box-selected-background);
 }
+
+.preset-batch-selected {
+    border: 2px solid #007bff;
+    box-shadow: 0 0 5px rgba(0, 123, 255, 0.5);
+}
+
+.batch-selection-checkbox {
+    position: absolute;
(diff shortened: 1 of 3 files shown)


bman654:
IMO this is a dangerous operation and it should start with everything UNCHECKED.  Maybe add a "select/unselect all" toggle to quickly flip, but I dont want to accidentally lose all my presets because they all start selected.

FurkanGozukara:
@bman654 it asks verification but certainly can be made that way. However @mcmonkey4eva has no intent to merge so i dont see any reason to make the changes as you recommended.

bman654:
FWIW I can't speak to code style issues, but one problem I see with correctness is that it is selecting all presets, even the ones not currently in the filtered view.  That seems problematic since it means it will be deleting presets that the user isn't even viewing.

FurkanGozukara:
@bman654 that is easy to fix / change. he doesnt like code style

mcmonkey4eva:
As explained on Discord, LLM vibe-coded PRs are not helpful, do not submit them

Issue #444, opened 2024-11-28
No custom nodes loading for me through SwarmUI


yo1nkers:
### Expected Behavior

I should expect to be able to use custom nodes. I have downloaded a watermarking workflow that utilizes custom nodes

### Actual Behavior

Does not work

### Steps to Reproduce

Completely unsure. From what I guess, just download SwarmUI and own it for half a year lol

### Debug Logs

Pulling latest changes...
Already up to date.
21:31:42.502 [Init] === SwarmUI v0.9.3.2 Starting at 2024-11-28 21:31:42 ===
21:31:42.559 [Init] Prepping extension: ImgMetadataConverter.ImgMetadataConverter...
21:31:42.562 [Init] Prepping extension: SwarmUI.Builtin_ScorersExtension.ScorersExtension...
21:31:42.562 [Init] Prepping extension: SwarmUI.Builtin_ImageBatchToolExtension.ImageBatchToolExtension...
21:31:42.563 [Init] Prepping extension: SwarmUI.Builtin_GridGeneratorExtension.GridGeneratorExtension...
21:31:42.563 [Init] Prepping extension: SwarmUI.Builtin_DynamicThresholding.DynamicThresholdingExtension...
21:31:42.564 [Init] Prepping extension: SwarmUI.Builtin_ComfyUIBackend.ComfyUIBackendExtension...
21:31:42.564 [Init] Prepping extension: SwarmUI.Builtin_AutoWebUIExtension.AutoWebUIBackendExtension...
21:31:42.598 [Init] Parsing command line...
21:31:42.599 [Init] Loading settings file...
21:31:42.615 [Init] Re-saving settings file...
21:31:42.634 [Init] Applying command line settings...
21:31:42.661 [Init] Prepping options...
21:31:42.679 [Init] Current git commit marked as date 2024-11-28 09:32:40
21:31:42.761 [Init] [ImgMetadataConverter] loaded
21:31:42.788 [Init] CPU Cores: 16 | RAM: 15.87 GiB total, 9.35 GiB available, 47.95 GiB total page file, 31.50 GiB available page file
21:31:42.789 [Init] GPU 0: NVIDIA GeForce RTX 3060 | Temp 56C | Util 1% GPU, 1% Memory | VRAM 12.00 GiB total, 10.77 GiB free, 1.06 GiB used
21:31:42.801 [Init] Loading models list...
21:31:43.447 [Init] Loading backends...
[...]

PR #997, opened 2025-08-31
Improved themability

mrblomblo:
- Moved nearly all hardcoded colors to their own vars or appropriate existing vars
- Made btn-secondary themeable
- Made toasts themeable

A few of the changes done cause a slight (almost unnoticeable) change in the default color of a small amount of things on Modern Dark, and a slightly more noticeable change on Modern Light. But makes all themes more consistent and more themed.

Notable changes observed in existing themes can be seen in the following two examples. IDK if this is considered breakage, as it (in my opinion) improves them:
_Example of language menu dropdown with the Punked theme (Legacy):_
image
_Example of btn-secondary with the Swarm Punk theme:_
image

Note that many variable names are long and descriptive (the longest one being `--model-block-menu-button-border-color-hover`).

*Edit: Removed irrelevant section & added one item to the list*

### DIFF
--- a/src/wwwroot/css/genpage.css
+++ b/src/wwwroot/css/genpage.css
@@ -468,8 +468,8 @@ body {
     opacity: 1.0;
     background-color: var(--emphasis) !important;
     cursor: pointer;
-    border: 1px solid var(--model-block-menu-button-border-color-hover);
-    color: var(--model-block-menu-button-color-hover);
+    border: 1px solid var(--button-border);
+    color: var(--text-strong);
 }
 .model-selected {
     border: 1px solid var(--box-selected-border-stronger);
--- a/src/wwwroot/css/genpage.css
+++ b/src/wwwroot/css/genpage.css
@@ -416,7 +416,7 @@ body {
     overflow: auto;
 }
 .model-block img:hover {
-    border: 2px solid white;
+    border: 2px solid var(--model-img-border-color-hover);
 }
 .model-block-small {
     height: 8rem;
@@ -468,7 +468,8 @@ body {
     opacity: 1.0;
     background-color: var(--emphasis) !important;
     cursor: pointer;
-    border: 1px solid black;
+    border: 1px solid var(--model-block-menu-button-border-color-hover);
+    color: var(--model-block-menu-button-color-hover);
 }
 .model-selected {
     border: 1px solid var(--box-selected-border-stronger);
@@ -590,10 +591,10 @@ body {
     margin-left: 5%;
 }
 .modal_error_bottom {
-    color: #ff0000;
+    color: var(--red);
 }
 .modal_success_bottom {
-    color: #00ff00;
+    color: var(--green);
 }
 #add_preset_modal {
     max-height: calc(95vh - var(--bs-modal-margin));
@@ -634,7 +635,7 @@ body {
     display: inline-block;
     color: var(--text);
     font-weight: bold;
-    border: 1px solid red;
+    border: 1px solid var(--red);
     border-radius: 0.5rem;
     margin-left: 0.2rem;
     padding-left: 0.2rem;
@@ -642,7 +643,7 @@ body {
     cursor: pointer;
 }
 .preset-remove-button:hover {
-    background-color: color-mix(in srgb, red 70%, black);
+    background-color: color-mix(in srgb, var(--red) 70%, black);
 }
 .new_preset_image {
     align-items: center;
@@ -720,8 +721,8 @@ body {
     padding-left: 2rem;
 }
 .setting-looks-wrong {
-    border: 1px  solid red !important;
-    box-shadow: 0px 0px 1px 1px red !important;
+    border: 1px  solid var(--red) !important;
+    box-shadow: 0px 0px 1px 1px var(--red) !important;
 }
 .group-label {
     position: relative;
@@ -1053,12 +1054,12 @@ body {
 .alt-prompt-image-container:hover .alt-prompt-image-container-remove-button {
     opacity: 1.0;
     background-color: color-mix(in srgb, transparent 50%, var(--background));
-    border: 1px solid red;
+    border: 1px solid var(--red);
 }
 
 .alt-prompt-image-container-remove-button:hover {
     opacity: 1.0;
-    background-color: color-mix(in srgb, red 70%, black) !important;
+    background-color: color-mix(in srgb, var(--red) 70%, black) !important;
     border: 1px solid black;
     cursor: pointer;
 }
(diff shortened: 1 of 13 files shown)


src/wwwroot/css/site.css
```diff
@@ -554,6 +597,14 @@ textarea, input, select {
     border-radius: 0.5rem;
     background-color: transparent;
 }
+.dropdown-menu {
+    --bs-dropdown-bg: var(--popup-back);
+    --bs-dropdown-border-color: gray;
+    color: var(--popup-front);
+}
+.dropdown-divider {
+    border-color: gray;
```
mcmonkey4eva: ?

src/wwwroot/css/site.css
```diff
@@ -530,12 +564,21 @@ textarea, input, select {
...
 }
 .basic-button:disabled {
-    background-color: color-mix(in srgb, var(--button-background) 40%, rgba(100, 0, 0, 0.5));
+    background-color: color-mix(in srgb, var(--button-background) 40%, var(--basic-button-bg-color-disabled));
     color: color-mix(in srgb, var(--button-text) 60%, transparent);
 }
 .simpler-button-disable:disabled {
     background-color: color-mix(in srgb, var(--button-background) 40%, transparent);
 }
+.btn-secondary {
```
mcmonkey4eva: thiissss should not be here, rather usages of bootstrap button classes should be fixed to normal classes like basic-button or a new equivalent one

src/wwwroot/css/site.css
```diff
@@ -13,7 +13,38 @@
...
+    --backend-loading-border-color: #335500;
+    --backend-waiting-border-color: #335500;
+    --backend-edit-btn-text-color-hover: #000000;
+    --backend-edit-btn-border-color-hover: #000000;
+    --backend-edit-btn-text-color-disabled: #505050;
+    --backend-edit-btn-border-color-disabled: #505050;
+    --backend-delete-btn-text-color-hover: #000000;
+    --backend-delete-btn-border-color-hover: #000000;
+    --backend-save-btn-text-color-hover: #000000;
+    --backend-save-btn-border-color-hover: #000000;
```
mcmonkey4eva: these are aggressively specific, and do not at all need to be.

src/wwwroot/css/genpage.css
```diff
@@ -590,10 +591,10 @@ body {
     margin-left: 5%;
 }
 .modal_error_bottom {
-    color: #ff0000;
+    color: var(--red);
 }
 .modal_success_bottom {
-    color: #00ff00;
+    color: var(--green);
```
mcmonkey4eva, line 597: These color-name vars are incredibly non-specific, and... I don't think even defined in site.css root?

src/wwwroot/css/site.css
mrblomblo: It makes the divider in the dropdown have the same color as the dropdown's outline, like it is originally.  
See images below:

_original:_
<details>
image
</details>  

_with gray dropdown divider:_
<details>
image
</details>  

_without gray dropdown divider:_
<details>
image
</details>

Though, maybe it is better to have different colors for them, unlike how it originally is?

src/wwwroot/css/genpage.css
mrblomblo: They are indeed not defined in site.css, but every theme has them defined (apart from modern_dark, modern_light, and solarized, which have them defined in modern.css). I could add them to the site.css root though, but that would require adding an `!important` to existing themes that modify those colors (especially Beweish, which is the only one that uses a different shade of green and red from the default shades).

What would be the best course of action here in your opinion? Revert those changes or add to site.css and add `!important` to Beweish's green and red vars (or just create new vars for those two instead)?

Edit: The backend edit, save, and delete buttons in site.css already use the `--green` and `--red` vars, so I assumed it wouldn't cause any issues

src/wwwroot/css/site.css
mcmonkey4eva: But... for a PR that's all about moving colors to variables, why are you adding brand-new raw css colors not in vars?

mrblomblo: Good point, that was a big oversight on my part lol. Will fix

src/wwwroot/css/site.css
mrblomblo: So if I added a `rst-welcome-msg-btn` class to the two buttons I intended on theming, and themed that class with `.btn-secondary.rst-welcome-msg-btn` (or maybe I don't need the `.btn-secondary`) then it'd be fine, or did I misinterpret?

I tried adding existing button classes to the two buttons I intended on theming, but it didn't work out.

src/wwwroot/css/genpage.css
mcmonkey4eva: Oh, oops, I got mixed up -- your PR adds all the new colors to site.css so I just assumed that's where they all were, I forgot most are defined theme-by-theme. That's messy, I like your system of base definitions in site.css better anyway.
I've amended the load order to let you define vars in site.css and override in a theme as you would expect to be able, without the `!important` hack.

src/wwwroot/css/site.css
```diff
@@ -402,6 +397,14 @@ textarea, input, select {
     animation-name: toast-flash;
     animation-duration: 2s;
 }
+.toast {
+    background-color: var(--toast-body-bg, var(--bs-toast-bg));
```
mrblomblo: I didn't find another way of making toasts themeable without breaking automatic dark/light mode switching. The problem with this approach is that the toast-theming variables can, to my knowledge, not be in the `:root` where all the other variables are, making it slightly less convenient for future theme creators.

mcmonkey4eva: Why can't they be in root? 0.o

mrblomblo: When I assigned `--toast-body-bg` the value `var(--bs-toast-bg)` in root, it just fully stopped working. And when I assigned it the color that `--bs-toast-bg` has (which I believe is `#444` or `#303030` for dark mode), it stopped automatically switching to the light-mode color when switching to a light-mode theme.

Therefore, I chose this approach that makes it themeable while still preserving the original functionality and colors for themes that choose not to theme it.

If I instead switch to using just `--toast-body-bg`, then I would assume that it'd need to be added to the modern_light and the legacy light-mode themes so they display the correct colors.

I could, of course, add the two toast color vars to the root but commented out, so future theme creators know about them.

Here's how it looks like _without_ `, var(--bs-toast-bg)` and `, var(--bs-toast-header-bg)`:
<details>
With the toast color vars defined in root with the colors that the BS vars use:

```css
    --toast-body-bg: #444;
    --toast-header-bg: #303030;
```

_Dark-mode (Correct colors):_
image

_Light-mode (Wrong colors):_
image

With the toast color vars defined in root with the BS vars themselves:

```css
    --toast-body-bg: var(--bs-toast-bg);
    --toast-header-bg: var(--bs-toast-header-bg);
```

image
</details>

Also, keeping the current  
```css
.toast-header {
    background-color: var(--toast-header-bg, var(--bs-toast-header-bg));
    color: var(--text);
}
```
implementation __and__ defining the toast vars in the root will still lead to the same problems, as it overrides the `, var(--bs-toast-header-bg)` that keeps the light/dark-mode switching working.

mcmonkey4eva: I mean, yes, put the relevant colors in the relevant theme files

mrblomblo:
Toasts are now properly stylable. Though, future light-mode themes will need to style the toast colors themselves as it doesn't switch to the light-mode colors automatically anymore (not a problem for existing light-mode themes, as I added the default light-mode colors for toasts to them).

Checkboxes are now styled with the emphasis color:
image
image

Text inputs and Textarea inputs now have their outline and border styled with the emphasis color:
image
image

PR #1002, opened 2025-09-02
Add GetGlobalStatus to API for getting system-wide status about running and waiting generations

suhuof:
Background:

I was working on a script which would use the SwarmUI API to fetch current generation count, and based on generation queue length and configured settings, would automatically search VastAI for GPU instances, setup and rent them, connect to them through secure SSH tunnel, and create backends for them using the SwarmUI API. They would also get automatically stopped and eventually destroyed if left unused for too long (configurable), along with the backends. The current SwarmUI API works well for managing the backends, but the GetCurrentStatus API call only returns the current session's generation counts, while the script would need the global counts to handle the backend scaling properly.

This PR:

I added GetGlobalStatus to AdminAPI for getting similar stats as GetCurrentStatus, but summed across all sessions. Also I ran the API doc regeneration and updated the doc files. In addition to the method I added, the other API docs were outdated too, so I pushed the updates to them in a separate commit. I can remove the other API doc update commit from the PR if you want to update them seperately.

This was just a quick way to have a working solution to test my script with, and it does work well with the script. But we can also discuss other ways to implement global generation status information to the API if you'd like a different solution. The script still has some bugs and needs some work before I can release it publicly, but already successfully automatically rents and configures instances for use with SwarmUI and generating with the instances is working.

### DIFF
--- a/docs/APIRoutes/AdminAPI.md
+++ b/docs/APIRoutes/AdminAPI.md
@@ -16,6 +16,7 @@ Administrative APIs related to server management.
 - HTTP Route [ChangeServerSettings](#http-route-apichangeserversettings)
 - HTTP Route [DebugGenDocs](#http-route-apidebuggendocs)
 - HTTP Route [DebugLanguageAdd](#http-route-apidebuglanguageadd)
+- HTTP Route [GetGlobalStatus](#http-route-apigetglobalstatus)
 - HTTP Route [GetServerResourceInfo](#http-route-apigetserverresourceinfo)
 - HTTP Route [InstallExtension](#http-route-apiinstallextension)
 - HTTP Route [ListConnectedUsers](#http-route-apilistconnectedusers)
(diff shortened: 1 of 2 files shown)


mcmonkey4eva:
remove the doc regen, that doesn't belong in a PR, that's a maintenance task

mcmonkey4eva:
also ftr backend autoscaling is on my near term todolist, and will definitely not work how you're doing it

suhuof:
Removed the API doc regen (except for the new method I added, or should I remove that too?). Good to hear that official autoscaling support is coming too, it would definitely be more user friendly to use directly from the UI. I'm trying to implement a quick script solution in the meanwhile :D. Though I learned that there is quite a lot of things that can go wrong (cuda errors with some instances, instance not being reachable with ssh after starting, outbid instances when using interruptible instances (which I wanted for cost savings), etc... What kind of plans do you have for the official autoscaling?

mcmonkey4eva:
See here for base structural plan https://github.com/mcmonkeyprojects/SwarmUI/blob/master/src/Backends/AutoScalingBackend.cs

suhuof:
> See here for base structural plan https://github.com/mcmonkeyprojects/SwarmUI/blob/master/src/Backends/AutoScalingBackend.cs

Looks good! Some ideas and thoughts:
- Is the extension check for start script needed? I think it could be useful to allow running e.g. python scripts directly here by specifying "python some.py" as the start command. This should work with the process spawning
- Maybe it would be useful to have separate stopped/destroyed status for backends? At least on vastai, stopped but non-destroyed ones could be useful to keep as quickly startable capacity which already have downloaded needed models on disk. When creating new instances, the model download can take a lot of time, especially on video models such as wan2.2. For me, it took over 10 minutes in some cases to download the model on a vastai instance, even when filtering out the instances with the slowest internet speeds. But this could make sense to also implement later in a future version.
- I think the authorization header is something that is randomly generated for each instance on vastai at least. Could be better to have the backend start script return the header if needed, similarly to how the backend address is now handled.
- I don't like the custom format for communication between SwarmUI and the backend start script. I think using something standard like json for passing data would be less error prone and easier to extend in the future.
- In stopping the instances, it would be bad if the stopping fails for some reason (internet connection is lost, etc.) and the instances on e.g. vastai are kept running and they would incur costs indefinitely, unless the user manually checks that nothing is left running there unused. Not sure what the best way to handle the stopping reliably would be, but maybe something like retrying the stopping indefinitely if it fails?

There is also another thing solving which would probably be useful for your implementation too. For my own implementation of the auto scaling, I was currently using ComfyUI instances on vastai, not SwarmUI ones, as the SwarmUI template on vastai needs manual setup to get a comfyui self-starting backend there. Since it is recommended to use SwarmUI as the remote backend, would it be possibly to make SwarmUI automatically create the self-starting backend without user intervention by editing the vastai template? It's easy to set the args passed to swarmui at least, when customizing the instance template. Also, if using vastai instances with multiple GPUs, would be nice to have it create one backend for each available GPU automatically on the vastai instance.

PR #1003, opened 2025-09-02
add stealth metadata for webp

Hugs288:


src/Utils/MetadataHelper.cs
```diff
@@ -29,6 +36,23 @@ public static void EncodeStealthMetadata(Image<Rgba32> image, string metadata, s
         }
     }
 
+    /// <summary>
+    /// Pre-processes the image to prevent the WebP encoder from discarding the alpha channel.
+    /// It does this by setting all alpha values to a slightly non-opaque value (254).
```
mcmonkey4eva: Surely *all* alpha values don't need to be set here? This is a weird hack.

ghost: you might be able to get away with less but the less pixels you set the more likely it is for the alpha layer to be discarded.

it is why stealth meta wasnt working with webp in the first place, the meta doesnt change the alpha layer enough for it to be not discarded.

i tried making the whole bottom row of pixels 254 but that was not enough, not sure of the exact threshold but it is required for a significant portion of the alpha layer to not be 255.

mcmonkey4eva: at a quick search, in webp lossless mode, https://github.com/SixLabors/ImageSharp/blob/main/src/ImageSharp/Formats/Webp/WebpCommonUtils.cs#L21 seems to be the root method, which if I'm not mistaken is an accelerated method that basically returns true if *even a single pixel* has any value that's not pure 255?

ghost: 🤷‍♂️neither lossy or lossless webp work without the hack or only a small amount of their pixels not being 255, ive tried.

its pretty obvious that its imagesharps fault because the same code with png works perfectly fine.

mcmonkey4eva: I did some testing and confirmed the odd behavior you saw, but found that a much more minimal altering of alpha values suffices to retain the alpha channel properly

mcmonkey4eva:
Noting here for your awareness: I finally took the time to look into webp metadata handling. As I suspected, it was indeed not nearly so simple as adding it to the ExtensionsWithMetadata, thanks to the existence of animated webps. It's now properly supported that any non-animated webp won't have any sidecar files and instead just trust the metadata in the file. Animated ones still use the sidecars.

ghost:
i do not use animated webp and you never mentioned them
image

PR #1012, opened 2025-09-06
Change redirect from 302 to 307

tgxn:
This will persist the request method.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/302

Causes issues where POST requests are redirected.
https://github.com/stavsap/comfyui-ollama/issues/83#issuecomment-3261557491


mcmonkey4eva:
waacky

ftr the upstream node's code is wrong, but ill post info on that there

Issue #952, opened 2025-08-11
UI tabs cut off when in Right Sidebar


bouchon1996:
### Feature Idea

There are several UI issues when the tabs are located on the right sidebar. For instance under Presets the Create Preset, Import Preset etc. buttons do not wrap down and are cut off unless given enough width to appear. Similarly, the filter and sort boxes for other tabs like Image History and Lora are cut off if the right column is too narrow to fit them. Ideally these buttons and boxes would wrap down to allow for all of the information to appear in the column. Some pictures are attached showing the issues. I believe all of these issues affect the mobile ui as well.

<img width="4556" height="1546" alt="Image" src="https://github.com/user-attachments/assets/d772934a-a399-4fef-8927-46eb8557f281" />

<img width="4556" height="1598" alt="Image" src="https://github.com/user-attachments/assets/aa2c2764-f514-4f42-bce0-d63d20df3353" />

mrblomblo:
Created a PR for the model browser header bar (https://github.com/mcmonkeyprojects/SwarmUI/pull/1015), though not implemented the same way you suggested

bouchon1996:
Awesome!> Created a PR for the model browser header bar (https://github.com/mcmonkeyprojects/SwarmUI/pull/1015), though not implemented the same way you suggested

PR #1015, opened 2025-09-06
Scrollable model browser header bar

mrblomblo:
For https://github.com/mcmonkeyprojects/SwarmUI/issues/952, though not implemented in the way suggested in the issue.

---

__Behavior before (not scrollable, as you already know):__
image

__Behavior after (scrollable):__  
_Firefox_

https://github.com/user-attachments/assets/537f98d5-96b2-42d8-a846-68762d956971

_Chromium_

https://github.com/user-attachments/assets/6c7a2301-8bad-4941-9f2d-30cafd61728c

---

The main difference between Firefox and Chromium (and their forks, I assume) is that the scrollbar does not hide itself on Chromium, which it does on Firefox. On Chromium, it also seems to make the header bar slightly taller to accommodate the new scrollbar, which it doesn't in Firefox (presumably due to it auto-hiding).

### DIFF
--- a/src/wwwroot/css/genpage.css
+++ b/src/wwwroot/css/genpage.css
@@ -778,7 +778,7 @@ body {
     width: calc(100% - 1rem);
     height: 2.5rem;
     display: block;
-    overflow-x: scroll;
+    overflow-x: auto;
     overflow-y: hidden;
     margin-left: 0.5rem;
 }
--- a/src/wwwroot/css/genpage.css
+++ b/src/wwwroot/css/genpage.css
@@ -778,7 +778,8 @@ body {
     width: calc(100% - 1rem);
     height: 2.5rem;
     display: block;
-    overflow: hidden;
+    overflow-x: scroll;
+    overflow-y: hidden;
     margin-left: 0.5rem;
 }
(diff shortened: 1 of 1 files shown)


src/wwwroot/css/genpage.css
```diff
@@ -778,7 +778,8 @@ body {
     width: calc(100% - 1rem);
     height: 2.5rem;
     display: block;
-    overflow: hidden;
+    overflow-x: scroll;
```
mcmonkey4eva: should probably be `auto`?

mrblomblo: I didn't notice any real difference between `auto` and `scroll`, but I can absolutely change it to `auto`

Edit: Ah, for consistency with the other `overflow-x` and `overflow-y`?

mcmonkey4eva: `auto` hides the scrollbar when there's not a need for it, `scroll` locks it on

mrblomblo: Ohhh so that's why it wasn't disappearing on Chromium I guess then?

PR #1021, opened 2025-09-08
add copy file button

Hugs288:
i am not sure if the linux and osx variations work, gemini wrote them and i can only test on windows.

### DIFF
--- a/src/WebAPI/T2IAPI.cs
+++ b/src/WebAPI/T2IAPI.cs
@@ -679,7 +679,7 @@ public static class T2IAPI
         }
         if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
         {
-            Process.Start("powershell.exe", $"-command Set-Clipboard -Path \"{Path.GetFullPath(path)}\"");
+            Process.Start("powershell.exe", $"-command Set-Clipboard -Path \\\"{Path.GetFullPath(path)}\\\"");
         }
         else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
         {
--- a/src/WebAPI/T2IAPI.cs
+++ b/src/WebAPI/T2IAPI.cs
@@ -31,6 +31,7 @@ public static class T2IAPI
         API.RegisterAPICall(ListImages, false, Permissions.ViewImageHistory);
         API.RegisterAPICall(ToggleImageStarred, true, Permissions.UserStarImages);
         API.RegisterAPICall(OpenImageFolder, true, Permissions.LocalImageFolder);
+        API.RegisterAPICall(CopyImagePath, true, Permissions.LocalImageFolder);
         API.RegisterAPICall(DeleteImage, true, Permissions.UserDeleteImage);
         API.RegisterAPICall(ListT2IParams, false, Permissions.FundamentalGenerateTabAccess);
         API.RegisterAPICall(TriggerRefresh, true, Permissions.FundamentalGenerateTabAccess); // Intentionally weird perm here: internal check for readonly vs true refresh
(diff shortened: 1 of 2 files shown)


mrblomblo:
I am on EndeavourOS (Arch-based Linux distro), kernel 6.16.4.

The copy button does not seem to work.  
Without xclip installed, I get the following error in the Swarm console:
```
18:51:43.715 [Error] [WebAPI] Error handling API request '/API/CopyImagePath' for user 'local': Internal exception: System.ComponentModel.Win32Exception (2): An error occurred trying to start process 'xclip' with working directory '/home/blomblo/Downloads/test/SwarmUI'. No such file or directory
   at System.Diagnostics.Process.ForkAndExecProcess(ProcessStartInfo startInfo, String resolvedFilename, String[] argv, String[] envp, String cwd, Boolean setCredentials, UInt32 userId, UInt32 groupId, UInt32[] groups, Int32& stdinFd, Int32& stdoutFd, Int32& stderrFd, Boolean usesTerminal, Boolean throwOnNoExec)
   at System.Diagnostics.Process.StartCore(ProcessStartInfo startInfo)
   at System.Diagnostics.Process.Start(ProcessStartInfo startInfo)
   at SwarmUI.WebAPI.T2IAPI.CopyImagePath(Session session, String path) in /home/blomblo/Downloads/test/SwarmUI/src/WebAPI/T2IAPI.cs:line 686
   at SwarmUI.WebAPI.API.HandleAsyncRequest(HttpContext context) in /home/blomblo/Downloads/test/SwarmUI/src/WebAPI/API.cs:line 148
```

And with xclip installed (`yay -S xclip`), I get the following in my clipboard:
```
�PNG

```

Which, when pasted into an image editor shows the following:
image

AshtakaOOf:
I think it'd be simpler to actually just copy the file path, since you could paste the file with it.

ghost:
i have no fucking idea how to deal with this, ill remove it for now.

if mcmonkey knows he can do it himself

mcmonkey4eva:
Copying should be done through the browser JS, not the server C#

bman654:
yeah its 1 google search away and about 5 lines of JavaScript code to copy an image to clipboard.

ghost:
copying webp through browser js is not supported

also png metadata is not preserved when copied with browser js (not sure about exif)

that would make this button useless, the whole point of this button is to make the open in folder -> right click image -> copy routine obsolete

mrblomblo:
I may be missing something, but isn't it possible to just right-click the image in the web browser? Or does that not keep the metadata intact or something?

ghost:
yes, thats effectively the same thing as copying it with browser js

the file always gets converted to png and its metadata is not always copied

its better to just copy the file directly

bman654:
maybe you are doing it wrong or your information is out of date.  Or the particular image element you've tried has itself stripped the metadata for some reason.

I just opened discord in my browser, found an image with metadata, clicked on it to "zoom" in.  Right-clicked -> Copy Image.  Pasted it into the message prompt, posted the message, ran prompt inspector on it and got back all of the metadata.  The image in question was a png.

Browsers also support copy/pasting webp because I've done that too.

mcmonkey4eva:
Yeah, that^

I've never had an issue with right click -> copy image preserving everything, at least in Firefox with regular png images.

ghost:
consider that everyone doesnt use png or firefox?

and no, chrome doesnt support copying webp images to clipboard, i tried

bman654:
my test was with chrome.

I just copied this webp (https://www.gstatic.com/webp/gallery/1.webp) and pasted here.  Looks like it got converted to png during the paste, but the "copy webp to clipboard" definitely worked.

image

Your information just does not seem up to date.  Maybe it would help if you gave a specific example that does not work instead of speaking in generalities (which are clearly wrong).

ghost:
no? it doesnt work. any webp you copy with chrome gets converted to png, because chrome doesnt support copying webp

ghost:
@mcmonkey4eva youre seriously not gonna merge this just because this issue doesnt affect you specifically?

mcmonkey4eva:
This is an incorrect and greatly overcomplicated solution to an issue that only exists if you go out of your way to cause it.

Anything related to copying an image needs to be handled in-browser, not in the C# server.

The best solutions here are your choice of:
- (1) don't go out of your way to use a file format you know has extremely limited compatibility, while expecting compatibility with things. How many different webp-related issues have you had to workaround or add special handling code for thus far?
- (2) figure out why your browser is unable to copy webp data and fix that on the browser end
- (3) use Firefox (or any other browser that doesn't have that same issue). Copies webp just fine, no conversion.
- (4) find a javascript based copying method that will bypass whatever chrome does re webp files which I presume is probably possible, js apis for this stuff have a lot in em

The terrible solution that I don't recommend at all but I suspect you'll do anyway based on the usual aggro attitude, is keep your current method and implement it as an Extension rather than core so you can keep using it as-is despite it not being merged to core.

ghost:
1) if you had actually bothered to read my comments you would know that this is not a webp specific issue. ANY image you copy with chrome and edge (and as far as i know any other chromium based browser) gets converted to png, loses all non-stealth metadata and usually makes the file significantly larger.

which is obviously a huge downgrade to just copying the file directly.

2) this issue is not specific to my browser

3) you would seriously rather tell me to switch browsers rather than merge a pr that makes the ui much more usable on every chromium based browser? which has 70%+ marketshare?

4) i believe the solution i prd to be the best as it bypasses the browser entirely

also i want to add that swarmui had by far the most broken implementation of webp i have ever seen in any software, im sorry for trying to fix it? i dont get it

PR #1022, opened 2025-09-08
Catppuccin themes

mrblomblo:
Added the four Catppuccin themes:

- Mocha
  image

- Macchiato
  image

- Frappé
  image

- Latte
  image

They are selectable in the installer.

__Licenses & Sources__

- Background image (ctp_background.jpg) [MIT]: https://github.com/zhichaoh/catppuccin-wallpapers/blob/main/patterns/line_icons.png
- Palette [MIT]: https://github.com/catppuccin/palette

---

I also made `border-danger` stylable, which I had forgotten to do in the Improved themability PR (https://github.com/mcmonkeyprojects/SwarmUI/pull/997). I also added an `!important` to the generate button's color when hovered in modern.css, as it didn't have any effect otherwise.

### DIFF
--- a/src/wwwroot/css/installer.css
+++ b/src/wwwroot/css/installer.css
@@ -123,8 +123,8 @@
 }
 .theme_preview_ctp_mocha .button_example {
     border: 1px solid #313244;
-    background-color: #fab387;
-    color: #1e1e2e;
+    background-color: #585b70;
+    color: #cdd6f4;
 }
 .theme_preview_ctp_macchiato {
     background-color: #181926;
@@ -132,8 +132,8 @@
 }
 .theme_preview_ctp_macchiato .button_example {
     border: 1px solid #363a4f;
-    background-color: #f5a97f;
-    color: #24273a;
+    background-color: #5b6078;
+    color: #cad3f5;
 }
 .theme_preview_ctp_frappe {
     background-color: #232634;
@@ -141,8 +141,8 @@
 }
 .theme_preview_ctp_frappe .button_example {
     border: 1px solid #414559;
-    background-color: #ef9f76;
-    color: #303446;
+    background-color: #626880;
+    color: #c6d0f5;
 }
 .theme_preview_ctp_latte {
     background-color: #dce0e8;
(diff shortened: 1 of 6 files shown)


AshtakaOOf:
Catppuccin does have multiple emphasis colors instead of just the orange.
Maybe it could be interesting to have an emphasis color override in the user settings ?
Also (just a nitpick) it would be nice if you could add missing themes to the installer.

mrblomblo:
Yeah, that would be cool. I have no idea how I'd implement the pick-your-own emphasis color, though...

The missing themes will come to the installer in a separate PR ;)

mcmonkey4eva:
A lot of the text/background color choices violate accessibility standards for minimum contrast ratio, eg these buttons:
image

Not the end of the day as not every theme needs to be super contrasty, but I have excellent vision and I still feel like the world's fading away trying to look at a lot of elements, on Mocha and Macchiato especially.

I don't think that's intended? Looking at the Catpuccin website, they seem to use colorful and/or high-contrast pairings for buttons similar to the element in the example test above
image
image

Glancing at their palette page, it looks like you gave button-text "Base" instead of "Text" colors?

mrblomblo:
oops, will fix once I get home

mrblomblo:
I fixed the button contrast issues and adjusted the installer page button text colors accordingly. I also changed the button background colors in the installer page to better represent a regular button (was using the emphasis color previously).  
Furthermore, I made emphasis buttons (primary buttons) have the emphasis text color for text in the button.

_Button contrast for Mocha theme:_
image

mrblomblo:
> Also (just a nitpick) it would be nice if you could add missing themes to the installer.
> @AshtakaOOf 

https://github.com/mcmonkeyprojects/SwarmUI/pull/1031

mcmonkey4eva:
iirc the color in the install page is meant to be the Generate button (arguably that whole mini-demo should be expanded/replaced with something a bit more useful, but, eh, mostly relying on users clicking it to see it properly atm)

mcmonkey4eva:
ngl I would still never use these themes myself as the low-contrast hurts my eyes still, but once the install page bit is fixed it's good enough to merge

Actually ill merge now since you have the separate install page PR and we can figure out any changes when merge-fixing that one

Issue #1027, opened 2025-09-10
TAESD previews not working in SwarmUI


Dwedit:
### Expected Behavior

In previous versions of everything (Swarm, Comfy, Embedded Python, Torch, etc), I had TAESD previews working in SwarmUI.  For some reason, they're not working anymore.

### Actual Behavior

If you try to enable TAESD previews, you get the error: "RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!

### Steps to Reproduce

* Started from a fresh installation of SwarmUI and the Comfy that it installs.
* Install the models, etc.

In Comfy, there are two ways to enable TAESD previews:
* Add "--preview-method taesd" as a command line argument 
  * (In Swarm's Backend configuration, adding "--preview-method taesd" to the command line requires you to also uncheck the "Enable Previews" box, as that checkbox adds "--preview-method latent2rgb" to the command line)

OR:
* Install the "ComfyUI-Manager" extension, click the "Manager" button, and change the preview mode to TAESD.

Either of these two ways to enable TAESD previews will replicate the problem in Swarm.

### Debug Logs

```
2025-09-10 11:28:03.669 [Init] === SwarmUI v0.9.7.0 Starting at 2025-09-10 11:28:03 ===
2025-09-10 11:28:03.856 [Init] Prepping extension: SwarmUI.Builtin_ScorersExtension.ScorersExtension...
2025-09-10 11:28:03.873 [Init] Prepping extension: SwarmUI.Builtin_ImageBatchToolExtension.ImageBatchToolExtension...
2025-09-10 11:28:03.874 [Init] Prepping extension: SwarmUI.Builtin_GridGeneratorExtension.GridGeneratorExtension...
2025-09-10 11:28:03.874 [Init] Prepping extension: SwarmUI.Builtin_DynamicThresholding.DynamicThresholdingExtension...
2025-09-10 11:28:03.874 [Init] Prepping extension: SwarmUI.Builtin_ComfyUIBackend.ComfyUIBackendExtension...
2025-09-10 11:28:03.880 [Init] Prepping extension: SwarmUI.Builtin_AutoWebUIExtension.AutoWebUIBackendExtension...
2025-09-10 11:28:03.989 [Init] Parsing command line...
2025-09-10 11:28:03.990 [Init] Loading settings file...
2025-09-10 11:28:04.014 [Init] Re-saving settings file...
[...]

PR #1029, opened 2025-09-10
Fix for TAESD previews, and option to select TAESD from backends screen

Dwedit:
For #1027
In "src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmKSampler.py"
Don't move the tensors to the CPU if the preview method is TAESD.  Prevents getting the error `"RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!"`

Also make it a lot easier to enable TAESD previews.  Previously, you had to either manually modify the command line, or you had to install Comfy UI Manager and change the preview method from within the Comfy Workflow tab.  Now it's available at the Backends screen.

### DIFF
--- a/src/BuiltinExtensions/ComfyUIBackend/ComfyUISelfStartBackend.cs
+++ b/src/BuiltinExtensions/ComfyUIBackend/ComfyUISelfStartBackend.cs
@@ -36,9 +36,9 @@ public class ComfyUISelfStartBackend : ComfyUIAPIAbstractBackend
         [ManualSettingsOptions(Impl = null, Vals = ["Latest", "None", "LatestSwarmValidated", "Legacy"], ManualNames = ["Latest", "None", "Latest Swarm Validated", "Legacy (Pre Sept 2024)"])]
         public string FrontendVersion = "LatestSwarmValidated";
 
-        [ConfigComment("Whether Comfy should generate image previews. If disabled, previews will not be generated, and images won't show up until they're done.\nLatent2RGB runs at 1/8 scale and does not match the final image's colors.\nTAESD (Tiny Auto Encoder for Stable Diffusion) is a distilled version of the Stable Diffusion VAE which runs much faster than the full VAE.")]
-        [ManualSettingsOptions(Impl = null, Vals = ["false", "true", "taesd"], ManualNames = ["No previews (fastest)", "Latent2RGB (fast)", "TAESD (high quality, less fast)"])]
-        public string EnablePreviews = "true";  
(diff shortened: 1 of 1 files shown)


src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmKSampler.py
```diff
@@ -76,7 +77,7 @@ def make_swarm_sampler_callback(steps, device, model, previews):
     def callback(step, x0, x, total_steps):
         pbar.update_absolute(step + 1, total_steps, None)
         if previewer:
-            if step == 0 or (step < 3 and x0.ndim == 5 and x0.shape[1] > 8):
+            if (step == 0 or (step < 3 and x0.ndim == 5 and x0.shape[1] > 8)) and args.preview_method != LatentPreviewMethod.TAESD:
```
mcmonkey4eva: i think there's a way to check via `previewer`, rather than CLI args

src/BuiltinExtensions/ComfyUIBackend/ComfyUISelfStartBackend.cs
```diff
@@ -39,6 +39,10 @@ public class ComfyUISelfStartSettings : AutoConfiguration
         [ConfigComment("If checked, tells Comfy to generate image previews. If unchecked, previews will not be generated, and images won't show up until they're done.")]
         public bool EnablePreviews = true;
 
+        [ConfigComment("Which type of image previews to use.  Using a higher quality method will increase image generation time.  TAESD (Tiny Auto Encoder for Stable Diffusion) is a distilled version of the Stable Diffusion VAE which runs much more quickly than the full VAE.")]
+        [ManualSettingsOptions(Impl = null, Vals = ["latent2rgb", "taesd"], ManualNames = ["Latent2RGB (fast)", "TAESD (higher quality, less fast)"])]
+        public string PreviewMethod = "latent2rgb";
```
mcmonkey4eva: Rather than a separate option, should adapt EnablePreviews, see eg how AutoUpdate works

Dwedit: I just really wanted to make sure that if a user had EnablePreviews turned off, that it wouldn't turn on.  The setting is off by default on AMD cards.  I don't know what happens if you turn on the setting with an AMD card, whether you get an exception, or just bad performance, but if previews are off, there's probably a good reason for it.

I'm not sure how to migrate a 'bool' setting to a 'string' setting.  We want a prior setting of "false" => "none", and "true" => "latent2rgb".

And even though TAESD looks a whole lot better than latent2rgb, I wouldn't want to turn it on by default due to performance concerns.  I just did some test gens, time was 24.5s with TAESD, and 22.75s with latent2rgb.

mcmonkey4eva: Don't rename the settings. Again, refer to AutoUpdate. It's an example case where a setting previously was a boolean and now isn't

Dwedit:
The mentioned changes were made.

PR #1031, opened 2025-09-11
Added missing themes and fixed theme order in installer

mrblomblo:
Added the three missing themes to the installer (Solarized Light, Swarm Punk, and Beweish), and moved Gravity Blue to match the order that the themes have in the theme selector in the User Settings.

image

### DIFF
--- a/src/Pages/Install.cshtml
+++ b/src/Pages/Install.cshtml
@@ -191,6 +191,7 @@
             <div class="form-check theme_preview_beweish" id="theme_input_beweish">
                 <input class="form-check-input" type="radio" name="theme_radio" id="theme_beweish" value="beweish">
                 <label class="form-check-label" for="theme_beweish">Beweish</label>
+                <span class="button_example">Example</span>
             </div>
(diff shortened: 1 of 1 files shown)


mrblomblo:
> iirc the color in the install page is meant to be the Generate button (arguably that whole mini-demo should be expanded/replaced with something a bit more useful, but, eh, mostly relying on users clicking it to see it properly atm)
> https://github.com/mcmonkeyprojects/SwarmUI/pull/1022#issuecomment-3283996020

Should I adjust all example button colors to match their respective generate button colors? For example, Modern Dark doesn't match the color of its generate button.
image
image

AshtakaOOf:
I think like you suggested [here](https://discord.com/channels/1243166023859961988/1243185862234210389/1415699466412888164) on Discord the idea of two buttons is better as it gives more context.

mcmonkey4eva:
tbh per discord discussion i'm currently of a mind to just ignore the details here until rewritten properly, dejanking the current setup won't do a ton as it's jank to the core.

Proper replacement being something like iframes that contain actual example UI elements and genuinely apply the stylesheet within each frame.

Issue #744, opened 2025-04-27
Feature Request: Add a border around the currently displayed image from the batch sidebar/image history


coredumperror:
### Feature Idea

When I'm clicking around in the batch sidebar to check out the images I just gen'd, or in the image history to find a pic I want to refine or reuse prompts from, it's really easy to lose track of which thumbnail is currently displayed at full size. If there was like a red border around the thumbnail associated with the image that's currently displayed in the main panel, that'd be really handy.

It wouldn't even have to track the current batch if that's too complicated. Just putting a border around the most recently clicked thumbnail (or the one that was most recently navigated to with the arrow keys) would be awesome.

CreativeAlloy:
Currently working on this one & coordinating with McMonkey to ensure nothing breaks, haha.

coredumperror:
That's great!

mrblomblo:
Jtreminio has created a draft PR for (what looks to be) this exact feature: https://github.com/mcmonkeyprojects/SwarmUI/pull/1032

mcmonkey4eva:
duplicate of https://github.com/mcmonkeyprojects/SwarmUI/issues/685

Issue #685, opened 2025-04-02
Make Image History follow and highlight the currently selected/viewed image


Taruvi:
### Feature Idea

Generally, I'll have a concept in mind and then generate a whole lot of images based on that concept. When I feel like I have enough of what I want, I'll stop and see what I'd like to work on some more. When I'm looking through Image History, I quite often scroll through with the arrow keys until I find something I might want to IMG2IMG. At this point, I do whatever it is I want, then I'll generally want to continue from where I left off and look back through some more to see what else I might like to work on.

At this point I might have scrolled through only a few or maybe dozens of images. I then have to spend the time to find the image I've just been working with so that I can continue scrolling through the rest. I'm stuck between manually spotting often quite similar images or trying to look for the matching filename. Spotting the exact image is not easy if there are many similar, especially when the thumbnails are tiny. The filename is not easy, either, depending on the filename format chosen. I use dates first which all blend into one and are hard to spot since the font is also tiny.

This can be mitigated somewhat by starring images during generation so there could potentially be fewer that you'd have to focus on. But sometimes I either forget or so many are so good that they end up all being starred, negating this approach.

Plus, if you're viewing the images in full screen, you can easily arrow through enough images that Image History is way off in terms of where you were in the list. If you have 1,000 images and you click on the very first one at the top, scroll through maybe 500 of them them with the arrow keys and then press escape, Image History will still be stuck at the very top still showing the first row.

So my feature request is that when arrowing through Image History, have the currently selected image highlighted, scrolling up or down the thumbnails as needed. In effect, have Image History follow your selection as you scroll through images.

Also, in Init Image in the sidebar, there is a button for 'Choose File' and beside it is the filename of the current file. This filename is what I sometimes use to try to find the image again in Image History. It would be very handy for this filename to possibly be clickable to jump to the image in Image History and highlight it with a single click. At the moment, clicking on the filename opens the 'Choose File' dialogue, when a dedicated button for that already exists right beside it.

jtreminio:
Implemented in #1121

Taruvi:
Indeed, it is. Thanks.

PR #1032, opened 2025-09-11
Highlight current selected image, move to next/previous on delete

jtreminio:
~This is a draft while I gather feedback.~

Ready for review.

---

# What this does

* Adds an _outline_ to the currently selected image in the `#current_image_batch` container
  * outline was chosen as it does not affect surrounding elements like a border does
* On no previous image, highlights the first image
* On image delete/remove from batch view, automatically highlights the _next_ image. If no next image exists, highlights the _previous_ image. If no images exist, noop
* Works with Image History

# What remains

* ✅ ~Updating all themes with new `.image-block-current` class~
* Implementing feedback regarding color choice, outline thickness, etc
* ✅ ~Making this work with Image History~

https://github.com/user-attachments/assets/c0925e78-66b4-408f-9474-29748a747c1b

Image History:

https://github.com/user-attachments/assets/eaab14e8-b390-4bc9-9e0d-bb58f549a064

### DIFF
--- a/src/wwwroot/css/themes/modern.css
+++ b/src/wwwroot/css/themes/modern.css
@@ -499,9 +499,6 @@
     background: rgba(0, 0, 0, 0.2) !important;
     border: 1px solid var(--box-selected-border-stronger);
 }
-.image-block.image-block-current {
-    outline: 3px solid var(--box-selected-border-stronger);
-}
 
 .model-block {
     border-radius: 10px;
--- a/src/wwwroot/css/themes/modern.css
+++ b/src/wwwroot/css/themes/modern.css
@@ -500,7 +500,7 @@
     border: 1px solid var(--box-selected-border-stronger);
 }
 .image-block.image-block-current {
-    outline: 3px solid color-mix(in srgb, var(--emphasis) 70%, white);
+    outline: 3px solid var(--box-selected-border-stronger);
 }
 
 .model-block {
--- a/src/wwwroot/css/themes/modern.css
+++ b/src/wwwroot/css/themes/modern.css
@@ -499,6 +499,9 @@
     background: rgba(0, 0, 0, 0.2) !important;
     border: 1px solid var(--box-selected-border-stronger);
 }
+.image-block.image-block-current {
+    outline: 3px solid color-mix(in srgb, var(--emphasis) 70%, white);
+}
 
 .model-block {
     border-radius: 10px;
--- a/src/wwwroot/js/genpage/gentab/currentimagehandler.js
+++ b/src/wwwroot/js/genpage/gentab/currentimagehandler.js
@@ -308,7 +308,6 @@ function clickImageInBatch(div) {
 
 /** Removes a preview thumbnail and highlights either next or previous image. */
 function removeImageBlockFromBatch(div) {
-    /** Current image isn't the highlighted one, so just remove it. */
     if (!div.classList.contains('image-block-current')) {
         div.remove();
         return;
@@ -908,7 +907,7 @@ function setCurrentImage(src, metadata = '', batchId = '', previewGrow = false, 
         curImg.appendChild(extrasWrapper);
     }
 
-    /** If switching the main image, we want to update the image preview  as well. */
+    // If switching the main image, we want to update the image preview  as well.
     let batchContainer = getRequiredElementById('current_image_batch');
     if (batchContainer) {
         let batchImg = batchContainer.querySelector(`[data-src="${src}"]`);
--- a/src/wwwroot/js/genpage/gentab/currentimagehandler.js
+++ b/src/wwwroot/js/genpage/gentab/currentimagehandler.js
@@ -306,6 +306,23 @@ function clickImageInBatch(div) {
     setCurrentImage(div.dataset.src, div.dataset.metadata, div.dataset.batch_id ?? '', imgElem && imgElem.dataset.previewGrow == 'true', false, true, div.dataset.is_placeholder == 'true');
 }
 
+// Removes a batch thumbnail and reassigns the 'image-block-current' class appropriately within its container
+function removeImageBlockFromBatch(div) {
+    if (!div.classList.contains('image-block-current')) {
+        div.remove();
+        return;
+    }
+
+    let chosen = div.previousElementSibling || div.nextElementSibling;
+    div.remove();
+
+    if (!chosen) {
+        return;
+    }
+
+    setCurrentImage(chosen.dataset.src, chosen.dataset.metadata, chosen.dataset.batch_id, true);
+}
+
 function rightClickImageInBatch(e, div) {
     if (e.shiftKey || e.ctrlKey) {
         return;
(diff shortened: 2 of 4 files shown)


Taruvi:
I know this is being mentioned in connection to [this](https://github.com/mcmonkeyprojects/SwarmUI/issues/744) feature request, but that's actually a duplicate of [my older feature request](https://github.com/mcmonkeyprojects/SwarmUI/issues/685).

In my request I also mention linking it to the Init Image preview as well. I don't know if that would be feasible.

Either way, this would be super useful if implemented as-is.

jtreminio:
@mcmonkey4eva Can I get feedback on

* what [CSS definition I should use for the outline](https://github.com/mcmonkeyprojects/SwarmUI/pull/1032/files#diff-49893c6aebd4a2d248727c1e974e09e0c91794c3f8bc680935763f0d7e8e454eR502-R504)?
* Do you think it's worth implementing the same outline applied to the Image History images?
* A follow-up PR would include not closing the large image modal when clicking Delete, and instead using the same next/previous logic as the non-modal large image container. Or, do you prefer that functionality added to this PR?

src/wwwroot/js/genpage/gentab/currentimagehandler.js
```diff
@@ -306,6 +306,20 @@ function clickImageInBatch(div) {
     setCurrentImage(div.dataset.src, div.dataset.metadata, div.dataset.batch_id ?? '', imgElem && imgElem.dataset.previewGrow == 'true', false, true, div.dataset.is_placeholder == 'true');
 }
 
+/** Removes a preview thumbnail and highlights either next or previous image. */
+function removeImageBlockFromBatch(div) {
+    /** Current image isn't the highlighted one, so just remove it. */
```
mcmonkey4eva: Comment isn't needed in the first place, but especially not a doc-comment `/**`

src/wwwroot/js/genpage/gentab/currentimagehandler.js
```diff
@@ -893,6 +907,19 @@ function setCurrentImage(src, metadata = '', batchId = '', previewGrow = false,
         curImg.appendChild(img);
         curImg.appendChild(extrasWrapper);
     }
+
+    /** If switching the main image, we want to update the image preview  as well. */
```
mcmonkey4eva: same as above, shouldn't be a doc comment in the middle of a function

src/wwwroot/js/genpage/gentab/currentimagehandler.js
```diff
@@ -893,6 +907,19 @@ function setCurrentImage(src, metadata = '', batchId = '', previewGrow = false,
...
+    let batchContainer = getRequiredElementById('current_image_batch');
+    if (batchContainer) {
+        let batchImg = batchContainer.querySelector(`[data-src="${src}"]`);
+        [...batchContainer.getElementsByClassName('image-block')].forEach(i => {
```
mcmonkey4eva: this is wacky syntax

jtreminio: There's no real package.json in the project, there's the jsconfig.json that has esnext as target, but I don't think that's an official "this project requires this min version" directive. The code as-is is the safest, best-supported way of iterating over the `HTMLCollection` returned by `getElementsByClassName`.

mcmonkey4eva: What? I'm not saying there's a bug, I'm saying the choice of syntax is wacky.
You used `===` where `==` is preferred, and you used `arr.forEach` where `for (x of arr)` is preferred.

jtreminio: ~`for (x of arr)` is blocking, `forEach` is async. Confirm you want the switch?~

For JS the preferred method for comparisons should always default to `===` unless you're ok with truthy checks. I'll change this to `==` in keep your preference in mind for future PRs.

src/wwwroot/js/genpage/gentab/currentimagehandler.js
```diff
@@ -893,6 +906,19 @@ function setCurrentImage(src, metadata = '', batchId = '', previewGrow = false,
...
     }
+
+    // If switching the main image, we want to update the image preview  as well.
+    let batchContainer = getRequiredElementById('current_image_batch');
+    if (batchContainer) {
+        let batchImg = batchContainer.querySelector(`[data-src="${src}"]`);
+        for (const i of batchContainer.getElementsByClassName('image-block')) {
+            if (batchImg && batchImg == i) {
+                i.classList.add('image-block-current');
+            } else {
```
mcmonkey4eva: The root of the prior comments was: match the existing code format when submitting a PR to any project. The format on this line does not match.

src/wwwroot/js/genpage/gentab/currentimagehandler.js
```diff
@@ -893,6 +906,19 @@ function setCurrentImage(src, metadata = '', batchId = '', previewGrow = false,
         curImg.appendChild(img);
         curImg.appendChild(extrasWrapper);
     }
+
+    // If switching the main image, we want to update the image preview  as well.
```
mcmonkey4eva: this comment should not be here

src/wwwroot/js/genpage/gentab/currentimagehandler.js
jtreminio: Is there a style guide I can import into IDE for auto-formatting? If not, are you OK with my creating a PR to add an eslint or .editorconfig file to help out with these things? They're not really things most engineers will notice unless it's super obvious, like 4-spaces.

mcmonkey4eva: Config for it is fine as long as it matches existing code properly

src/wwwroot/css/themes/modern.css
```diff
@@ -499,6 +499,9 @@
     background: rgba(0, 0, 0, 0.2) !important;
     border: 1px solid var(--box-selected-border-stronger);
 }
+.image-block.image-block-current {
```
mcmonkey4eva: `image-block-current` should be defined in genpage, not modern

src/wwwroot/js/genpage/gentab/currentimagehandler.js
bman654: FWIW I have a branch where I worked on an editorconfig but never got past the c# style b/c in practice the style seems to have changed over time, so a single config couldn't match all files especially with respect to when newlines are applied around various types of `{}`.  I never tried to just do one for the JS code.

I also noticed the source is a mad mix of line endings, text encodings, and BOM's.  Would be really nice if we could do a PR that cleaned that up and added a `.gitattributes` file to enforce a standard.

mcmonkey4eva: C# and JS have different format standards, but should be internally consistent without themselves. No editor config should be applied to C#, the format there is strictly enforced by VS already.

All encoding should be UTF-8 without BOM, any exception to that was a mistake (if in C# files, Visual Studio sometimes wonks that for some reason). Line endings should be reliable iirc, git OS-normalizes them by default.

bman654: here's an example of C# formatting inconsistency.  Most places have`if` blocks with [newlines](https://github.com/mcmonkeyprojects/SwarmUI/blob/master/src/Text2Image/T2IParamTypes.cs#L1129).  Except when [they do not have newlines](https://github.com/mcmonkeyprojects/SwarmUI/blob/master/src/Text2Image/T2IParamTypes.cs#L1106).

mcmonkey4eva: Again, you don't need to worry about C# formatting. It's mostly IDE enforced and the special cases are intended.

PR #1036, opened 2025-09-13
[Installer] Dynamic theme preview loading, Iframe previews, and other small improvements

mrblomblo:
Added dynamic loading of theme previews, and switched the preview type to iframe. Other than that, I made some small aesthetic improvements such as: emphasis color on the border of the preview of the currently applied theme, better centered theme preview title, and a small space to the left of the radio.

image

_In the picture, you may notice that a theme or two doesn't have the correct colors on the generate button. I believe that is caused by the theme not explicitly defining said color in its theme file. I think I may know a fix for this issue._

### DIFF
--- a/src/Pages/Install.cshtml
+++ b/src/Pages/Install.cshtml
@@ -140,79 +140,20 @@
 
     <div id="installer_section_themes" style="display: none;">
         <div class="install_q_head">Choose a theme:</div>
-        <div class="install_hint">You can always change this later in the User Settings page.</div>
+        <div class="install_hint">You can always change this later in the User Settings page. The theme previews may take some time to load.</div>
         <br>
         <fieldset class="form-group theme_preview" id="theme_selection_field">
-            <div class="form-check theme_preview_modern_dark" id="theme_input_modern_dark">
-                <input class="form-check-input" type="radio" name="theme_radio" id="theme_moderndark" value="modern_dark" checked="">
-                <label class="form-check-label" for="theme_moderndark">Modern Dark</label>
-                <span class="button_example">Example</span>
-            </div>
-            <div class="form-check theme_preview_modern_light" id="theme_input_modern_light">
-                <input class="form-check-input" type="radio" name="theme_radio" id="theme_modernlight" value="modern_light">
-                <label class="form-check-label" for="theme_modernlight">Modern Light</label>
-                <span class="button_example">Example</span>
-            </div>
-            <div class="form-check theme_preview_solarized" id="theme_input_solarized">
-                <input class="form-check-input" type="radio" name="theme_radio" id="theme_solarized" value="solarized">
-                <label class="form-check-label" for="theme_solarized">Solarized Light</label>
-                <span class="button_example">Example</span>
-            </div>
-            <div class="form-check theme_preview_dark_dreams" id="theme_input_dark_dreams">
-                <input class="form-check-input" type="radio" name="theme_radio" id="theme_darkdreams" value="dark_dreams">
-                <label class="form-check-label" for="theme_darkdreams">Dark Dreams</label>
-                <span class="button_example">Example</span>
-            </div>
-            <div class="form-check theme_preview_gravity_blue" id="theme_input_gravity_blue">
-                <input class="form-check-input" type="radio" name="theme_radio" id="theme_gravityblue" value="gravity_blue">
-                <label class="form-check-label" for="theme_gravityblue">Gravity Blue</label>
-                <span class="button_example">Example</span>
-            </div>
-            <div class="form-check theme_preview_cyber_swarm" id="theme_input_cyber_swarm">
-                <input class="form-check-input" type="radio" name="theme_radio" id="theme_cyberswarm" value="cyber_swarm">
-                <label class="form-check-label" for="theme_cyberswarm">Cyber Swarm</label>
-                <span class="button_example">Example</span>
-            </div>
-            <div class="form-check theme_preview_punked" id="theme_input_punked">
-                <input class="form-check-input" type="radio" name="theme_radio" id="theme_punked" value="punked">
-                <label class="form-check-label" for="theme_punked">Punked</label>
-                <span class="button_example">Example</span>
-            </div>
-            <div class="form-check theme_preview_eyesear_white" id="theme_input_eyesear_white">
-                <input class="form-check-input" type="radio" name="theme_radio" id="theme_eyesearwhite" value="eyesear_white">
-                <label class="form-check-label" for="theme_eyesearwhite">Eyesear White</label>
-                <span class="button_example">Example</span>
-            </div>
-            <div class="form-check theme_preview_swarmpunk" id="theme_input_swarmpunk">
-                <input class="form-check-input" type="radio" name="theme_radio" id="theme_swarmpunk" value="swarmpunk">
-                <label class="form-check-label" for="theme_swarmpunk">Swarm Punk</label>
-                <span class="button_example">Example</span>
-            </div>
-            <div class="form-check theme_preview_beweish" id="theme_input_beweish">
-                <input class="form-check-input" type="radio" name="theme_radio" id="theme_beweish" value="beweish">
-                <label class="form-check-label" for="theme_beweish">Beweish</label>
-                <span class="button_example">Example</span>
-            </div>
-            <div class="form-check theme_preview_ctp_mocha" id="theme_input_ctp_mocha">
-                <input class="form-check-input" type="radio" name="theme_radio" id="theme_ctpmocha" value="ctp_mocha">
-                <label class="form-check-label" for="theme_ctpmocha">Catppuccin Mocha</label>
-                <span class="button_example">Example</span>
-            </div>
-            <div class="form-check theme_preview_ctp_macchiato" id="theme_input_ctp_macchiato">
-                <input class="form-check-input" type="radio" name="theme_radio" id="theme_ctpmacchiato" value="ctp_macchiato">
-                <label class="form-check-label" for="theme_ctpmacchiato">Catppuccin Macchiato</label>
-                <span class="button_example">Example</span>
-            </div>
-            <div class="form-check theme_preview_ctp_frappe" id="theme_input_ctp_frappe">
-                <input class="form-check-input" type="radio" name="theme_radio" id="theme_ctpfrappe" value="ctp_frappe">
-                <label class="form-check-label" for="theme_ctpfrappe">Catppuccin Frappé</label>
-                <span class="button_example">Example</span>
-            </div>
-            <div class="form-check theme_preview_ctp_latte" id="theme_input_ctp_latte">
-                <input class="form-check-input" type="radio" name="theme_radio" id="theme_ctplatte" value="ctp_latte">
-                <label class="form-check-label" for="theme_ctplatte">Catppuccin Latte</label>
-                <span class="button_example">Example</span>
-            </div>
+            @foreach (var theme in Program.Web.RegisteredThemes.Values)
+            {
+                string themeName = $"{theme.Name.Replace(" (Legacy)","")}";
+                string themeId = $"{theme.ID}";
+                string themeIdNoUnderscore = $"{theme.ID.Replace("_", "")}";
+                <div class="form-check theme_preview_@themeId" id="theme_input_@themeId">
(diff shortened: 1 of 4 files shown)


src/Pages/Install.cshtml
```diff
@@ -140,79 +140,20 @@
 
     <div id="installer_section_themes" style="display: none;">
         <div class="install_q_head">Choose a theme:</div>
-        <div class="install_hint">You can always change this later in the User Settings page.</div>
+        <div class="install_hint">You can always change this later in the User Settings page. The theme previews may take some time to load.</div>
```
mcmonkey4eva: This notice shouldn't be needed I'd think?

src/Pages/Install.cshtml
```diff
@@ -140,79 +140,20 @@
...
-            </div>
+            @foreach (var theme in Program.Web.RegisteredThemes.Values)
+            {
+                string themeName = $"{theme.Name.Replace(" (Legacy)","")}";
+                string themeId = $"{theme.ID}";
+                string themeIdNoUnderscore = $"{theme.ID.Replace("_", "")}";
+                <div class="form-check theme_preview_@themeId" id="theme_input_@themeId">
+                    <input class="form-check-input" type="radio" name="theme_radio" id="theme_@themeIdNoUnderscore" value="@themeId" @(themeId == "modern_dark" ? "checked" : "")>
+                    <label class="form-check-label" for="theme_@themeIdNoUnderscore">@themeName</label>
+                    <iframe class="form-check-iframe" src="ThemePreview?theme=@themeId" loading="lazy"></iframe>
```
mcmonkey4eva: drop the `loading="lazy"`

src/Pages/ThemePreview.cshtml
```diff
@@ -0,0 +1,34 @@
...
+    <link rel="stylesheet" href="css/genpage.css?vary=@Utilities.VaryID" />
+    <link rel="stylesheet" href="css/themepreview.css?vary=@Utilities.VaryID" />
+}
+<script>
+    (function () {
+        const urlParams = new URLSearchParams(window.location.search);
+        const themeId = urlParams.get('theme');
+
+        if (themeId) {
+            const linkElement = document.createElement('link');
```
mcmonkey4eva: why is this javascript? just do it in C# in the Header section

mcmonkey4eva: or if using JS, call the existing JS set theme function

src/Pages/Install.cshtml
mrblomblo: Correct, I put it there when I noticed that they were taking like 5 seconds to load, not noticing that I had the browser devtools open :facepalm:

mrblomblo:
I fixed the things you brought up, but I still don't get why the iframes are inheriting the variable values from the parent page. It even seems to have gotten worse now after setting the theme with C# for some reason. I've already tried the following to fix it, but none of my attempts have even been slightly successful:
- Add `contain: layout style paint;` and `isolation: isolate;` to the CSS for the iframe element in the parent page
- Add `sandbox="allow-same-origin allow-scripts"` to the iframe element in the parent page
- Change all variable values from their current values to both `unset` and `initial` before loading the theme both inside the iframe and the iframe element in the parent page, both with and without !important (i.e., try to reset them)
- Set `all: unset;` for :root in the iframe
- Change the CSS file-loading order in the iframe

mrblomblo:
I suppose I could test web components. If that doesn't work, then I will probably close this PR and create a new one with just the dynamic theme preview loading (for a clean commit history) and leave the improved preview part to someone more capable than me.

bman654:
> I fixed the things you brought up, but I still don't get why the iframes are inheriting the variable values from the parent page. It even seems to have gotten worse now after setting the theme with C# for some reason. I've already tried the following to fix it, but none of my attempts have even been slightly successful:
> 
> * Add `contain: layout style paint;` and `isolation: isolate;` to the CSS for the iframe element in the parent page
> * Add `sandbox="allow-same-origin allow-scripts"` to the iframe element in the parent page
> * Change all variable values from their current values to both `unset` and `initial` before loading the theme both inside the iframe and the iframe element in the parent page, both with and without !important (i.e., try to reset them)
> * Set `all: unset;` for :root in the iframe
> * Change the CSS file-loading order in the iframe

what specifically is happening?

mrblomblo:
> > I fixed the things you brought up, but I still don't get why the iframes are inheriting the variable values from the parent page. It even seems to have gotten worse now after setting the theme with C# for some reason. I've already tried the following to fix it, but none of my attempts have even been slightly successful:
> > 
> > * Add `contain: layout style paint;` and `isolation: isolate;` to the CSS for the iframe element in the parent page
> > * Add `sandbox="allow-same-origin allow-scripts"` to the iframe element in the parent page
> > * Change all variable values from their current values to both `unset` and `initial` before loading the theme both inside the iframe and the iframe element in the parent page, both with and without !important (i.e., try to reset them)
> > * Set `all: unset;` for :root in the iframe
> > * Change the CSS file-loading order in the iframe
> 
> what specifically is happening?

Ah, sorry for not saying what's happening... Here's a screenshot of how it looks in its current state:
<details>
image
</details>

Basically, it seems like all theme preview iframes inherit the CSS variable values from the parent page (the install page) at load (they don't change styling when changing the theme preview), ignoring the variable values they are given by the theme file they're supposed to be showing off in the actual theme preview iframe. Well, apart from Punked, which seems to have a higher specificity for the generate button and middle generate button in its CSS file.

I hope I made sense lol

mcmonkey4eva:
more likely you're just seeing the default cookie-based theme application and not replacing it properly?

mrblomblo:
That could definitely be the issue, I hadn't thought of that

mrblomblo:
Also, do you think it'd be better to have a (Legacy) badge after legacy themes' names, similar to how it looks in the theme picker in the user settings, or leave it the way it is now? @mcmonkey4eva

mcmonkey4eva:
leave that as-is

src/Pages/Shared/_Layout.cshtml
```diff
@@ -6,8 +6,12 @@
     <title id="page_title">@ViewData["Title"] - SwarmUI - @Program.ServerSettings.UserAuthorization.InstanceTitle</title>
     @{
         string themeId = "modern_dark";
+        string pageThemeOverride = ViewContext.HttpContext.Request.Query["theme"].ToString();
+        if (!string.IsNullOrEmpty(pageThemeOverride) && Program.Web.RegisteredThemes.ContainsKey(pageThemeOverride)) {
```
mrblomblo: A side effect from this is that if the user adds `?theme=<theme-id>` to the URL, it changes the theme to whatever theme has that ID

mrblomblo:
Theme previews now work properly.
image

src/Pages/Shared/_Layout.cshtml
mcmonkey4eva: ideally this would be limited to only work on the ThemePreview page. Not sure the optimal way to do that, maaaybe just jank a `if (Page.EndsWith("/ThemePreview"))`

Or maybe set a `ViewData["theme_override"]` in ThemePreview based on the query, and read that value in Layout

mrblomblo: Yeah I was thinking that would be best too, but I wasn't sure of how to do it. Will see if I can do it with ViewData

mrblomblo:
I've implemented the ViewData thing you suggested. I am trying to find the Core Parameters so I can copy it over, but I can't find it

mcmonkey4eva:
Parameters on the generate tab are generated by JS. `makeSliderInput` and all those

mrblomblo:
I am getting nowhere. I cannot figure out how to get the params to be generated by JS, and I cannot find any code I can use as an example in any of the files.

mcmonkey4eva:
Search the function I named above in vs code?

image

mrblomblo:
Sorry, was having a slow day. I am making progress right now, but the Seed input buttons are acting a little weird:  
Modern theme:
image

Legacy theme:
image

<details>
<summary>Current code</summary>

```js
function reuseLastParamVal() {
    return null; // Prevent error in browser console
}

function generateInputs() {
    const container = getRequiredElementById('input_group_content_coreparameters');
    container.innerHTML = 
        makeNumberInput(null, 'input_images', 'images', 'Images', '', 1, 1, 10000, 1, 'big', false, true)
        + makeNumberInput(null, 'input_seed', 'seed', 'Seed', '', -1, -1, 9223372036854776000, 1, 'seed', false, true)
        + makeSliderInput(null, 'input_steps', 'steps', 'Steps', '', 20, 0, 500, 0, 100, 1, false, false, true)
        + makeSliderInput(null, 'input_cfgscale', 'cfgscale', 'CFG Scale', '', 7, 0, 100, 0, 20, 0.5, false, false, true);
    autoNumberWidth(document.getElementById('input_images'));
    enableSliderAbove(document.getElementById('input_steps'));
    enableSliderAbove(document.getElementById('input_cfgscale'));
}

document.addEventListener('DOMContentLoaded', function() {
    generateInputs();
});
```

</details>

Also, should I make the popover buttons functional, or is that a bit overkill?

mcmonkey4eva:
None of the buttons should do anything when clicked (other than basic buttonstate stuff)

mrblomblo:
Core params are here now, but two icons are present in the seed buttons in modern themes...
image

src/Pages/ThemePreview.cshtml
```diff
@@ -0,0 +1,55 @@
...
+        + makeNumberInput(null, 'input_seed', 'seed', 'Seed', '', -1, -1, 9223372036854776000, 1, 'seed', false, true)
+        + makeSliderInput(null, 'input_steps', 'steps', 'Steps', '', 20, 0, 500, 0, 100, 1, false, false, true)
+        + makeSliderInput(null, 'input_cfgscale', 'cfgscale', 'CFG Scale', '', 7, 0, 100, 0, 20, 0.5, false, false, true);
+    autoNumberWidth(document.getElementById('input_images'));
+    enableSliderAbove(document.getElementById('input_steps'));
+    enableSliderAbove(document.getElementById('input_cfgscale'));
+}
+
+document.addEventListener('DOMContentLoaded', function() {
+    makeGenericPopover('input_images', 'Images', 'Integer', 'How many images to generate at once.', ["1", "4"])
```
mcmonkey4eva: these shouldn't be here.

src/wwwroot/css/installer.css
```diff
@@ -24,166 +24,38 @@
...
+    width: 100%;
+    height: calc(100% - 2rem);
+    border-bottom-left-radius: calc(0.5rem - 1px);
+    border-bottom-right-radius: calc(0.5rem - 1px);
+    position: absolute;
+    bottom: 0;
+    left: 0;
 }
 label {
+    margin-top: 0.125rem;
```
mcmonkey4eva: These edits are modifying a lot more than just the theme preview stuff, probably shouldn't be here

src/wwwroot/css/themepreview.css
```diff
@@ -0,0 +1,33 @@
+html, body {
+    height: 100%
```
mcmonkey4eva: `;` at end

mcmonkey4eva:
`but two icons are present in the seed buttons in modern themes...`
That was caused by some jank css in modern.css - fixed now, and I merged master to this PR, it does indeed fix the issue here.

This is looking really solid. Just needs some minor refinements I think and it's ready to go.

src/Pages/ThemePreview.cshtml
mrblomblo: woops, forgot to remove them from my testing

mrblomblo:
I fixed the things you brought up.

What would you think of having the background of the previews use the same noise as the generate tab for extra immersion?
imageimage

mcmonkey4eva:
yeah the noise background texture should be included where relevant yeh

mrblomblo:
Noise image is now present in the core params section.
image

Btw, when I was doing this, I noticed that the noise image isn't applied to light themes. Is that intentional?

mrblomblo:
Oh yeah, in the browser console there is an error about recursion that keeps spamming the browser error log whenever you click in a preview. It stops spamming when you click outside the preview.

https://github.com/user-attachments/assets/f0669c71-a2ec-40f0-9881-dd5824edaef6

_(Sorry about the laggy recording)_

mcmonkey4eva:
Couldn't replicate that error, but regardless it should not be valid to click inside the box, I pushed a commit that makes it so if you click inside it just triggers the checkbox select instead.

PR #1040, opened 2025-09-14
Fix contrast issues with text color on emphasis background

mrblomblo:
Also added `!important` to `.basic-button.btn-primary`'s emphasis text because it wasn't applying to every btn-primary (such as the ones in the model metadata editor).

### DIFF
--- a/src/wwwroot/css/themes/cyber_swarm.css
+++ b/src/wwwroot/css/themes/cyber_swarm.css
@@ -28,6 +28,7 @@
     --background-danger: #200b0b;
     --shadow: #3f3f46;
     --emphasis: #00bc8c;
+    --emphasis-text: #ffffff;
     --light-border: #3f3f46;
     --red: #ff0000;
     --green: #00ff00;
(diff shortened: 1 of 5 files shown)


src/wwwroot/css/site.css
```diff
@@ -574,7 +574,7 @@ textarea, input, select {
 }
 .basic-button.btn-primary {
     background-color: var(--emphasis);
-    color: var(--emphasis-text);
+    color: var(--emphasis-text) !important;
```
mcmonkey4eva, line 577: I must've missed this being added in the previous PR - this var doesn't even exist on all themes

mrblomblo: oh, hmm. All the modern themes have it (since it's in modern_dark and modern_light, and some modify the var), and they've had it since before my PRs. I could add the `--emphasis-text` var to the legacy themes with appropriate colors if that is preferable

mcmonkey4eva: yes adding to the legacy themes is required to use it outside of the modern.css files

mrblomblo:
Added the `--emphasis-text` var to the legacy themes with appropriate colors.

Issue #1042, opened 2025-09-14
Full View of unfinished generation should update automatically as previews arrive


Dwedit:
### Feature Idea

Currently, the Full View (clicking on a picture to make it big) is frozen while previews are arriving, despite that the center panel's view (current_image) is automatically updating, and the right-side "Batch View" (current_image_batch) is also automatically updating.  I think the Full View should be automatically updating as well.

Dwedit:
There's also an issue with the AutoSwapImagesIncludesFullView feature, it affects which image shows up next when you use Left/Right arrow keys.  Rather than using the currently displayed big image, it uses the "current_image" (active in the center) instead.

Dwedit:
I plan on creating a PR from this: https://github.com/mcmonkeyprojects/SwarmUI/compare/master...Dwedit:SwarmUI:mine-image-previews3

PR #1046, opened 2025-09-15
Live previews for Full View image, bug fixes, and new mode for AutoSwap images

Dwedit:
This PR makes the following changes:

* Added live previews for Full View images.
* Fixed several bugs when opening a preview image:
  * There was a bug where you had to rapidly click on the Current Image or Batch Image in order to open the Full View.  Normally you click once to select the image, then click a second time to open the Full View.  But it was failing to open the Full View if a preview arrived while the image was selected, instead re-selecting the image.  Opening a Full View required you to click rapidly.
  * Opening the Full View opened a stale preview image rather than the most recent iteration of the preview
* Changes to prevent you from losing your place when navigating Full View images:
  * Previously, when a new image arrived, your movement cursor was reset back to the leftmost image.  This changes it so that your currently viewed image in Full View mode is used instead of the Current Image.
  * New feature: Option to only swap to new images in Full View if we are looking at the newest completed image.  This allows you to watch new images come in, but also move right and left without losing your place.  When you want new images again, move to either the newest completed gen, or the current preview.
* Also fixes an exception when you star an image, but you removed the star button via user settings.

Related issue: #1042

### DIFF
--- a/src/Core/Settings.cs
+++ b/src/Core/Settings.cs
@@ -409,8 +409,9 @@ public class Settings : AutoConfiguration
         [ConfigComment("If true, images in the main center area will always grow to better fill the screen.")]
         public bool CenterImageAlwaysGrow = false; // TODO: UserUI
 
-        [ConfigComment("If true, when 'Auto Swap To Images' is enabled, and you have FullView open, the FullView will also be swapped.\nIf false, the FullView will not change.")]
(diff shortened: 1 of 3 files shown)


mcmonkey4eva:
This is a lot at once - please submit key parts here separately

PR #1049, opened 2025-09-17
Add live previews to the Full View

Dwedit:
Add live previews to the Full View

In order to support live previews, the Full View must now keep track of the batch ID, so the function `ImageFullViewHelper.showImage` has a new parameter to pass in the batch ID.  This is saved as the `this.currentBatchId` member. The `img` element with id `imageview_popup_modal_img` is also exposed as `this.imgElement`.

With the img element exposed, and the Batch ID now associated with the Full View, the function `GenerateHandler.internalHandleData` can now safely update the Full View's img src when its batch ID matches.

---

I tested out that all calls to `showImage` to see if the batch id was correct.  But there was one I couldn't test, that's the `gotImageResult` function.  I'll have to assume that the `batchId` function parameter that's passed in is correct, since it's also passed to `setCurrentImage`.

### DIFF
--- a/src/wwwroot/js/genpage/gentab/currentimagehandler.js
+++ b/src/wwwroot/js/genpage/gentab/currentimagehandler.js
@@ -177,9 +177,10 @@ class ImageFullViewHelper {
         img.style.height = `${newHeight}%`;
     }
 
-    showImage(src, metadata) {
+    showImage(src, metadata, batchId) {
         this.currentSrc = src;
         this.currentMetadata = metadata;
+        this.currentBatchId = batchId;
         let isVideo = isVideoExt(src);
         let encodedSrc = escapeHtmlForUrl(src);
         let imgHtml = `<img class="imageview_popup_modal_img" id="imageview_popup_modal_img" style="cursor:grab;max-width:100%;object-fit:contain;" src="${encodedSrc}">`;
@@ -197,6 +198,7 @@ class ImageFullViewHelper {
                 ${formatMetadata(metadata)}
             </div>
         </div>`;
+        this.imgElement = document.getElementById("imageview_popup_modal_img");
         let subDiv = this.content.querySelector('.image_fullview_extra_buttons');
         for (let added of buttonsForImage(getImageFullSrc(src), src, metadata)) {
             if (added.href) {
@@ -300,7 +302,7 @@ function toggleSeparateBatches() {
 function clickImageInBatch(div) {
     let imgElem = div.getElementsByTagName('img')[0];
     if (currentImgSrc == div.dataset.src) {
-        imageFullView.showImage(div.dataset.src, div.dataset.metadata);
+        imageFullView.showImage(div.dataset.src, div.dataset.metadata, div.dataset.batch_id);
         return;
     }
     setCurrentImage(div.dataset.src, div.dataset.metadata, div.dataset.batch_id ?? '', imgElem && imgElem.dataset.previewGrow == 'true', false, true, div.dataset.is_placeholder == 'true');
@@ -447,7 +449,7 @@ function shiftToNextImagePreview(next = true, expand = false) {
         divs[newIndex].querySelector('img').click();
         if (expand) {
             divs[newIndex].querySelector('img').click();
-            imageFullView.showImage(currentImgSrc, currentMetadataVal);
+            imageFullView.showImage(currentImgSrc, currentMetadataVal, 'history');
             imageFullView.pasteState(expandedState);
         }
         return;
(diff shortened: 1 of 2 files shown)


src/wwwroot/js/genpage/helpers/generatehandler.js
```diff
@@ -199,6 +199,10 @@ class GenerateHandler {
                         }
                         this.setImageFor(imgHolder, data.gen_progress.preview);
                     }
+                    if (data.gen_progress.preview && imageFullView.isOpen() && imageFullView.imgElement && imageFullView.currentBatchId == thisBatchId)
+                    {
```
mcmonkey4eva: shouldn't be a newline here for JS code

src/wwwroot/js/genpage/gentab/currentimagehandler.js
```diff
@@ -447,7 +449,7 @@ function shiftToNextImagePreview(next = true, expand = false) {
         divs[newIndex].querySelector('img').click();
         if (expand) {
             divs[newIndex].querySelector('img').click();
-            imageFullView.showImage(currentImgSrc, currentMetadataVal);
+            imageFullView.showImage(currentImgSrc, currentMetadataVal, 'history');
```
mcmonkey4eva, line 451: this should probably be null

src/wwwroot/js/genpage/gentab/currentimagehandler.js
```diff
@@ -177,9 +177,10 @@ class ImageFullViewHelper {
         img.style.height = `${newHeight}%`;
     }
 
-    showImage(src, metadata) {
+    showImage(src, metadata, batchId) {
```
mcmonkey4eva: this should probably have a default `= null`

src/wwwroot/js/genpage/helpers/generatehandler.js
```diff
@@ -199,6 +199,10 @@ class GenerateHandler {
                         }
                         this.setImageFor(imgHolder, data.gen_progress.preview);
                     }
+                    if (data.gen_progress.preview && imageFullView.isOpen() && imageFullView.imgElement && imageFullView.currentBatchId == thisBatchId)
```
mcmonkey4eva: this block should be inside the block above

src/wwwroot/js/genpage/gentab/currentimagehandler.js
```diff
@@ -691,7 +693,9 @@ function setCurrentImage(src, metadata = '', batchId = '', previewGrow = false,
     img.dataset.src = src;
     img.dataset.metadata = metadata || '{}';
     img.dataset.batch_id = batchId;
-    img.onclick = () => imageFullView.showImage(img.dataset.src, img.dataset.metadata);
+    img.onclick = () => {
```
mcmonkey4eva: why was this reformatted? put it back

src/wwwroot/js/genpage/gentab/currentimagehandler.js
```diff
@@ -197,6 +198,7 @@ class ImageFullViewHelper {
                 ${formatMetadata(metadata)}
             </div>
         </div>`;
+        this.imgElement = document.getElementById("imageview_popup_modal_img");
```
mcmonkey4eva: obscure, but for html engine stability reasons, move this to after the modalJq show

src/wwwroot/js/genpage/helpers/generatehandler.js
Dwedit: I had the newline there so I could set breakpoints and verify that the arguments were good.  Do I need to change it back?

src/wwwroot/js/genpage/gentab/currentimagehandler.js
Dwedit: In the function `setCurrentImage`, there's this lambda:
`imageFullView.showImage(img.dataset.src, img.dataset.metadata, img.dataset.batch_id)`

It's used in two situations:
* When you click on an image from History twice.  `img.dataset.batch_id` is the word 'history' when that happens.
* When you click on the Current Image in the center panel.  The batch_id will either be a real batch id if you opened an image from the batch (right panel), or the word 'history'.

Then there's the code for `shiftToNextImagePreview`.
It is checking if the batch_id is the word 'history', and using that as a sign that you want to move back and forth through history rather than a batch.
To keep things consistent, I'm passing in 'history' as well because the word 'history' is already getting passed in by other code.

src/wwwroot/js/genpage/helpers/generatehandler.js
Dwedit: I would disagree here.  It's a check that exclusively applies to the Full View.  The block above is a check that only applies to the Current Image.  The only part that really applies to all three is the `data.gen_progress.preview` part.

src/wwwroot/js/genpage/gentab/currentimagehandler.js
Dwedit: It got reformatted so I could put breakpoints inside using the script debugger.  If it's a problem, I'll put it back.

mcmonkey4eva: yes put it back

src/wwwroot/js/genpage/helpers/generatehandler.js
mcmonkey4eva: That is... not correct.

PR #1063, opened 2025-09-25
Add diffusion_models support for model downloader

nerdylive123:


nerdylive123:
### Feature description
I'm trying to add an option where user can download "diffusion_models" directly from the model downloader utility page
image
I tested the fork, it works fine after downloading & refreshed model lists:
image

### Testing Other types of model downloading:

- Lora:

image
image
image

- VAE
image
image
image

PR #1077, opened 2025-10-06
allow file paths for creating input images

DSMann:
Adds a new function to create Image objects using a local file path for input, and adjusts the param validation to allow existing file paths through.

I am not a C# developer, so any feedback/considerations are very welcome.

https://discord.com/channels/1243166023859961988/1423175127377379368

### DIFF
--- a/src/Text2Image/T2IParamSet.cs
+++ b/src/Text2Image/T2IParamSet.cs
@@ -2,6 +2,7 @@
 using SwarmUI.Accounts;
 using SwarmUI.Core;
 using SwarmUI.Utils;
+using System.IO;
 
 namespace SwarmUI.Text2Image;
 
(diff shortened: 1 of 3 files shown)


mcmonkey4eva:
(1) raw file paths here are explicitly and intentionally not supported
(2) this does not actually solve the linked discord question regardless
(3) if you'd like to contribute to Swarm, that is great, but please start with the Issues marked with the "Easy PR" tag. First-time PRs should start by solving issues that are relatively simpler to solve so you can learn the basics before taking on more advanced challenges. The actual solution needed for this one is much more complex and non-obvious, it is not suitable for a first-time contrib.

DSMann:
Makes sense.  I'll take a look at some of those tags.  Beyond asking first (which I should have done), is there a way to know that paths aren't supposed to be supported there?

Issue #1097, opened 2025-10-24
Support Chroma Radiance


gitmylo:
### Feature Idea

Chroma radiance is an in-development diffusion model (with a few releases so far) which runs in pixel space rather than latent space.
To make chroma radiance work with swarmui, a new model type should be added for it, and most of the behavior should be the same as regular chroma, with 2 changes:
1. The vae should select "`pixel_space`", which is a new option for pixel space models, not an actual model so no downloads required
2. Instead of `EmptySD3LatentImage`, use `EmptyChromaRadianceLatentImage` for creating the "latent", or rather empty image
After those changes are made, the model should work without issues

Model: https://huggingface.co/lodestones/Chroma1-Radiance
Comfyui supports it: https://github.com/comfyanonymous/ComfyUI/pull/9682

PR #1100, opened 2025-10-25
Support Chroma radiance

gitmylo:
Support for https://huggingface.co/lodestones/Chroma1-Radiance

Basic chroma radiance support through adding a new model type (doesn't automatically get detected right now though). It uses the "pixel_space" vae option and the emptychromaradiancelatent node. Other than that there are no changes, confirmed working. I don't know what the optimal settings are, afaik the resolution doesn't really matter much.

For #1097

### DIFF
--- a/docs/Model Support.md
+++ b/docs/Model Support.md
@@ -309,7 +309,7 @@ These steps are not friendly to beginners (if Sana gains popularity, likely more
     - Add a Mask, draw a dot anywhere in the empty area (this is just a trick to tell the editor to automask all the empty area to the side, you don't need to mask it manually)
     - Type your prompt, hit generate
 
-### Chroma
+# Chroma
 
 - Chroma is a derivative of Flux, and is supported in SwarmUI
     - FP8 Scaled versions here: <https://huggingface.co/silveroxides/Chroma1-HD-fp8-scaled/tree/main>
@@ -333,6 +333,10 @@ These steps are not friendly to beginners (if Sana gains popularity, likely more
     - **Sigma Shift:** Defaults to `1`
     - **Resolution:** `1024x1024` or nearby values. The *HD* models were trained extra on `1152x1152`.
 
+# Chroma Radiance
+
+- TODO: fill me in
+    - Download here <https://huggingface.co/lodestones/Chroma1-Radiance/tree/main>
 
 # Lumina 2
 


src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
```diff
@@ -1099,7 +1106,7 @@ void doVaeLoader(string defaultVal, string compatClass, string knownName)
                     {
                         dtype = "default";
                     }
-                    else if (IsNvidiaCosmos2() || IsOmniGen() || IsChroma())
+                    else if (IsNvidiaCosmos2() || IsOmniGen() || IsChroma() || IsChromaRadiance())
```
mcmonkey4eva, line 1109: is this actually needed on radiance too?

src/Text2Image/T2IModelClassSorter.cs
```diff
@@ -449,6 +449,10 @@ JToken GetEmbeddingKey(JObject h)
         {
             return isChroma(h);
         }});
+        Register(new() { ID = "chroma-radiance", CompatClass = "chroma-radiance", Name = "Chroma Radiance", StandardWidth = 1024, StandardHeight = 1024, IsThisModelOfClass = (m, h) =>
+        {
+            return false;
```
mcmonkey4eva: should be autodetectable, and I doubt they properly annotate metadata

src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
gitmylo: I don't know, I added the IsChromaRadiance check to any place that previously checked for chroma, and changed the vae and empty latent nodes only

src/Text2Image/T2IModelClassSorter.cs
gitmylo: Should now be autodetectable, I had a gguf so I wasn't really able to test it, downloaded the model to confirm and it detected it correctly.

src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
gitmylo: After removing it and running the model, the outputs are fried and extremely oversaturated, so yes it does seem to be needed

src/Text2Image/T2IModelClassSorter.cs
mcmonkey4eva: ggufs should autodetect the same as safetensors

src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
```diff
@@ -1316,7 +1323,14 @@ void doVaeLoader(string defaultVal, string compatClass, string knownName)
                 ["shift"] = UserInput.Get(T2IParamTypes.SigmaShift, 1)
             });
             LoadingModel = [samplingNode, 0];
-            doVaeLoader(UserInput.SourceSession?.User?.Settings?.VAEs?.DefaultFluxVAE, "flux-1", "flux-ae");
+            if (IsChromaRadiance())
+            {
+                LoadingVAE = CreateVAELoader("pixel_space");
```
mcmonkey4eva, line 1328: This is wrong. Use `doVaeLoader`, and define the relevant VAE in `CommonModels`

gitmylo: I don't think that would make sense. There is no model to download so attempting to register it would yield an `InvalidOperationException`.

An alternative could be to modify `doVaeLoader` to allow `pixel_space` as an option, and possibly additionally to add `pixel_space` to the dropdown, although I doubt any other models can give sensible results with that.

I originally used the CreateVAELoader because it seemed more appropriate as pixel_space is not even a vae, instead it's an option that is always available in comfyui (at least in versions that support chroma radiance and up).

So should I leave it or edit `doVaeLoader` to allow `pixel_space` as an option, which would end up doing the same thing?

gitmylo: Actually, doing `doVaeLoader("pixel_space", null, null);` will work without issues.

mcmonkey4eva: Oh! I didn't realize it was a special option like that

Issue #1102, opened 2025-10-30
Bug Report: Edit Metadata not saving new Preview images


richardtallent:
### Expected Behavior

1. When a model/lora/etc. doesn't have a proper preview image and I load metadata from CivitAI, it should check the Image checkbox.

2. When the Image checkbox is checked and a preview image has been chosen or loaded from CivitAI, it should be saved to the metadata.

3. After saving, the image should be immediately replaced in the UI.

### Actual Behavior

I noticed over the past few weeks that when I edit the metadata for a new model, I couldn't get the thumbnail loaded from CivitAI to save.

- The checkbox wasn't auto-checking after loading from CivitAI, even though I previously didn't have a preview image set for that model.
- Even manually checking the checkbox didn't result in the preview image being saved.
- Once I fixed the problem of it not saving, it still didn't update the UI, I had to refresh the browser window.

### Steps to Reproduce

1. Copy a new model, LORA, etc. into the appropriate folder.
2. Go to Edit Metadata.
3. Load from CivitAI.
4. Check the Image checkbox.
5. Click Save.
6. Note that the image didn't save.
7. Reload the browser tab. Image still not saved, so it's not just a refresh issue.

### Debug Logs

I didn't need this, fixed it already, PR incoming.

mcmonkey4eva:
This was apparently a bug induced by recent caching optimizations, fixed

PR #1103, opened 2025-10-30
Fix model thumbnail not saving #1102

richardtallent:
This addresses the bug relayed in #1102, Thumbnail Not Saving

Model thumbnail images loaded from CivitAI were not being saved to metadata, leaving users with placeholder thumbnails even after successfully loading real model images.

Commit `6a007358` introduced a regression by disabling the auto-enabling behavior of the image toggle in the model editing interface.

This meant that when users loaded thumbnails from CivitAI, the toggle remained unchecked, preventing the images from being saved. Even manually checking the checkbox was not successful.

## Solution

1. **Restored Auto-Enable**: Automatically enables the image toggle when no existing preview image exists or the current preview is the placeholder.

2. **Added Immediate Cache Refresh**: After a successful save, the UI is updated with the new preview image.

## Other Changes

I noticed that this repository doesn't have a Prettier configuration. Many developers (myself included) use Prettier as our default JS formatter, so saving *any* files here leads to massive reformatting. I added a `.prettierignore` file to ensure that Prettier doesn't reformat files in this repo.

*(While Prettier is sometimes frustratingly opinionated, I recommend checking it out and creating a Prettier config file that works for you, it will make it much easier to ensure that contributors follow your preferences. Same with ESLint.)*

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,0 @@
-# Disable Prettier for all files in this repository
-**/*
--- a/src/wwwroot/js/genpage/gentab/models.js
+++ b/src/wwwroot/js/genpage/gentab/models.js
@@ -182,8 +182,7 @@ function editModel(model, browser) {
     let curImg = document.getElementById('current_image_img');
     if (curImg && curImg.tagName == 'IMG') {
         setImageFileDirect(modelsHelpers.imageElem, curImg.src, 'cur', 'cur', () => {
-            // Auto-enable if there's no existing preview image or it's the placeholder
-            modelsHelpers.enableImageElem.checked = !model.preview_image || model.preview_image == 'imgs/model_placeholder.jpg';
+            modelsHelpers.enableImageElem.checked = false;
             run();
         });
     }
@@ -236,9 +235,7 @@ function edit_model_load_civitai() {
         }
         if (img) {
             setImageFileDirect(modelsHelpers.imageElem, img, 'cur', 'cur', () => {
-                // Auto-enable if there's no existing preview image or it's the placeholder
-                let prev = curModelMenuModel ? curModelMenuModel.preview_image : null;
-                modelsHelpers.enableImageElem.checked = !prev || prev == 'imgs/model_placeholder.jpg';
+                modelsHelpers.enableImageElem.checked = false;
                 triggerChangeFor(modelsHelpers.enableImageElem);
             });
         }


.prettierignore
```diff
@@ -0,0 +1,2 @@
+# Disable Prettier for all files in this repository
+**/*
```
mcmonkey4eva: this should not be here

src/wwwroot/js/genpage/gentab/models.js
```diff
@@ -182,7 +182,8 @@ function editModel(model, browser) {
     let curImg = document.getElementById('current_image_img');
     if (curImg && curImg.tagName == 'IMG') {
         setImageFileDirect(modelsHelpers.imageElem, curImg.src, 'cur', 'cur', () => {
-            modelsHelpers.enableImageElem.checked = false;
+            // Auto-enable if there's no existing preview image or it's the placeholder
+            modelsHelpers.enableImageElem.checked = !model.preview_image || model.preview_image == 'imgs/model_placeholder.jpg';
```
mcmonkey4eva: nope that was changed for a reason

.prettierignore
richardtallent: There's no Prettier configuration in this repo. That means for many developers, their personal Prettier settings are going to blow up your formatting.

If you are actually using Prettier, can you add a prettier config file instead so our auto-saved format matches yours?

src/wwwroot/js/genpage/gentab/models.js
richardtallent: Doesn't make sense to me, but I'll take your word for it.

When I tested, even checking the Image checkbox after loading from CivitAI didn't save the preview. It seemed like perhaps the checkbox was getting unchecked during the Save process.

But that *may* have actually been a situation where it was being saved, but just didn't refresh on-screen (the other change I made, which was also a regression -- it used to update automatically).

I'll go back to the main branch and confirm, and revert this change if the cache refresh alone fixes the issue.

.prettierignore
mcmonkey4eva: I am not using that, and you should not have tools like that globally enabled by default - it should default off, and only be enabled for repos where you've chosen to enable it.

richardtallent: I was here to help fix a bug and try to make it easier to contribute, not get into an argument about tooling. I'll remove it and put it in my local folder without committing.

mcmonkey4eva:
Replaced with https://github.com/mcmonkeyprojects/SwarmUI/commit/aef918dfc6514aaa0f6444b8af04e7bab4128ea1

PR #1107, opened 2025-11-01
Adds DeleteImageCycleEnd flag to prevent cycling from ends when deleting image

jtreminio:
Flag ON (default behavior) depending on flag of existing `DeleteImageBehavior` flag:
* `[E, D, C, B, A◄]` Delete A
* `[E◄, D, C, B]` Jumps to E

or,

* `[E◄, D, C, B, A]` Delete E
* `[D, C, B, A◄]` Jumps to A

With flag OFF:
* `[E, D, C, B, A◄]` Delete A
* `[E, D, C, B◄]` Jumps to B

or,

* `[E◄, D, C, B, A]` Delete E
* `[D◄, C, B, A]` Jumps to D

This is separate from the behavior of the `ImageShiftingCycles` - that flag will now decide if you can shift from end to end with left/right arrows. This new `DeleteImageBehavior` flag decides if deleting applies the same behavior.

### DIFF
--- a/src/Core/Settings.cs
+++ b/src/Core/Settings.cs
@@ -390,6 +390,9 @@ public class Settings : AutoConfiguration
 
             [ConfigComment("If true, shifting to next/previous image (eg with arrow keys) in history or batch view,\ncycles at the ends (jumps from the start to the end or vice versa).\nIf false, shifting will simply stop at the ends.")]
             public bool ImageShiftingCycles = true;
+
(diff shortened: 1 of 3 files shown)


mcmonkey4eva:
replaced via https://github.com/mcmonkeyprojects/SwarmUI/commit/ffefd5b41656fe0b9cbd919b2394716a52ddf1b2

PR #1109, opened 2025-11-03
Feature/model parameter overrides

richardtallent:
As discussed in Discord, this allows users to select a preset for any model or LoRA. The preset is added when the model/LoRA are selected. These are stored as **user settings**.

For models, the preset is *unselected* (if not applied already) if the user clicks another model that has a different preset.

Documentation is included.

This adds two optional user settings as well:

- Whether to auto-apply presets when selecting a model/LoRA rather than just selecting the corresponding preset.
- Whether to ignore LoRA presets for LoRA loaded through a model's preset, if the LoRA is loaded with a weight of 0 (so the model can effectively disable the LoRA).

This is my first deep trip into the code, so please let me know if I've accidentally re-implemented utility functions, etc. that already exist. I've tried to follow the conventions and style of this repo as much as I can, and to minimize the "blast radius" of the new code.

### DIFF
--- a/src/WebAPI/BasicAPIFeatures.cs
+++ b/src/WebAPI/BasicAPIFeatures.cs
@@ -224,10 +224,10 @@ public static class BasicAPIFeatures
             },
             "model_preset_links": {
                 "Stable-Diffusion": {
-                    "modelnamehere": "preset_title"
+                    "modelnamehere": ["preset_title"]
                 },
                 "LoRA": {
-                    "modelnamehere": "preset_title"
+                    "modelnamehere": ["preset_title"]
                 }
             },
             "autocompletions": ["Word\nword\ntag\n3"]
@@ -262,7 +262,7 @@ public static class BasicAPIFeatures
 
     [API.APIDescription("Saves a reference to a preset for a model or LoRA to the user's data.", "\"success\": \"true\"")]
     public static async Task<JObject> SetPresetLinks(Session session,
-        [API.APIParameter("Send the raw data as eg 'LoRA': { 'Name': 'Preset' }, 'Stable-Diffusion': { ... }")] JObject raw)
+        [API.APIParameter("Send the raw data as eg 'LoRA': { 'Name': ['Preset'] }, 'Stable-Diffusion': { ... }")] JObject raw)
     {
         raw.Remove("session_id");
         session.User.SaveGenericData("modelpresetlinks", "full", raw.ToString(Formatting.None));
--- a/src/wwwroot/js/genpage/gentab/models.js
+++ b/src/wwwroot/js/genpage/gentab/models.js
@@ -744,11 +744,11 @@ class ModelBrowserWrapper {
                 interject += `${getOptLine("Default LoRA Weight", model.data.lora_default_weight)}${getOptLine("Default LoRA Confinement", confinementName)}`;
                 searchableAdded += `, Default LoRA Weight: ${model.data.lora_default_weight}, Default LoRA Confinement: ${confinementName}`;
             }
-            let linkedPreset = modelPresetLinkManager.getLink(this.subType, model.data.name);
-            if (linkedPreset) {
-                searchableAdded += `, Linked Preset: ${linkedPreset}`;
+            let linkedPresets = modelPresetLinkManager.getLinks(this.subType, model.data.name);
+            if (linkedPresets.length > 0) {
+                searchableAdded += `, Linked Presets: ${linkedPresets.join(', ')}`;
             }
-            description = `<span class="model_filename">${isStarred ? 'Starred: ' : ''}${escapeHtml(display)}</span><br>${getLine("Title", model.data.title)}${getOptLine("Author", model.data.author)}${getLine("Type", model.data.class)}${interject}${getOptLine('Trigger Phrase', model.data.trigger_phrase)}${getOptLine("Linked Preset", linkedPreset)}${getOptLine('Usage Hint', model.data.usage_hint)}${getLine("Description", model.data.description)}<br>`;
+            description = `<span class="model_filename">${isStarred ? 'Starred: ' : ''}${escapeHtml(display)}</span><br>${getLine("Title", model.data.title)}${getOptLine("Author", model.data.author)}${getLine("Type", model.data.class)}${interject}${getOptLine('Trigger Phrase', model.data.trigger_phrase)}${getOptLine("Linked Presets", linkedPresets.join(', '))}${getOptLine('Usage Hint', model.data.usage_hint)}${getLine("Description", model.data.description)}<br>`;
             let cleanForDetails = (val) => val == null ? '(Unset)' : safeHtmlOnly(val).replaceAll('<br>', '&emsp;');
             detail_list.push(cleanForDetails(model.data.title), cleanForDetails(model.data.class), cleanForDetails(model.data.usage_hint ?? model.data.trigger_phrase), cleanForDetails(model.data.description));
             if (model.data.local && permissions.hasPermission('edit_model_metadata')) {
@@ -1006,13 +1006,7 @@ function directSetModel(model) {
         modelName = name;
     }
     reviseBackendFeatureSet();
-    let presetTitle = modelPresetLinkManager.getLink('Stable-Diffusion', modelName);
-	if (presetTitle) {
-		let presetData = allPresetsUnsorted?.find(p => p.title == presetTitle);
-		if (presetData) {
-            presetHelpers.addPreset(presetData);
-		}
-	}
+    modelPresetLinkManager.addPresetsFrom('Stable-Diffusion', modelName);
     getRequiredElementById('input_model').dispatchEvent(new Event('change'));
     let aspect = document.getElementById('input_aspectratio');
     if (aspect) {
--- a/src/wwwroot/js/genpage/gentab/presets.js
+++ b/src/wwwroot/js/genpage/gentab/presets.js
@@ -56,28 +56,43 @@ class ModelPresetLinkManager {
         this.links = {};
     }
 
-    getLink(subtype, modelName) {
-        return this.links[subtype]?.[cleanModelName(modelName)] || null;
+    getLinks(subtype, modelName) {
+        return this.links[subtype]?.[cleanModelName(modelName)] || [];
     }
 
     setLink(subtype, modelName, presetTitle) {
+        // TODO: Static single set is silly. There should be a multi-select in the UI for this.
         modelName = cleanModelName(modelName);
         if (!presetTitle?.trim()) {
-            this.clearLink(subtype, modelName);
+            this.clearLinks(subtype, modelName);
             return;
         }
         this.links[subtype] ??= {};
-        if (this.links[subtype][modelName] != presetTitle) {
-            this.links[subtype][modelName] = presetTitle;
-            this.saveModelPresetLinks();
+        this.links[subtype][modelName] ??= [];
+        if (this.links[subtype][modelName].length != 1 || this.links[subtype][modelName][0] != presetTitle) {
+            this.links[subtype][modelName] = [presetTitle];
+            this.save();
         }
     }
 
-    clearLink(subtype, modelName) {
+    addLink(subtype, modelName, presetTitle) {
+        modelName = cleanModelName(modelName);
+        if (!presetTitle?.trim()) {
+            return;
+        }
+        this.links[subtype] ??= {};
+        this.links[subtype][modelName] ??= [];
+        if (!this.links[subtype][modelName].includes(presetTitle)) {
+            this.links[subtype][modelName].push(presetTitle);
+            this.save();
+        }
+    }
+
+    clearLinks(subtype, modelName) {
         modelName = cleanModelName(modelName);
         if (this.links[subtype]?.[modelName]) {
             delete this.links[subtype]?.[modelName];
-            this.saveModelPresetLinks();
+            this.save();
         }
     }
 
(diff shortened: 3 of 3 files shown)


src/wwwroot/js/genpage/gentab/models.js
```diff
@@ -645,7 +654,17 @@ class ModelBrowserWrapper {
                 interject += `${getOptLine("Default LoRA Weight", model.data.lora_default_weight)}${getOptLine("Default LoRA Confinement", confinementName)}`;
                 searchableAdded += `, Default LoRA Weight: ${model.data.lora_default_weight}, Default LoRA Confinement: ${confinementName}`;
             }
-            description = `<span class="model_filename">${isStarred ? 'Starred: ' : ''}${escapeHtml(display)}</span><br>${getLine("Title", model.data.title)}${getOptLine("Author", model.data.author)}${getLine("Type", model.data.class)}${interject}${getOptLine('Trigger Phrase', model.data.trigger_phrase)}${getOptLine('Usage Hint', model.data.usage_hint)}${getLine("Description", model.data.description)}<br>`;
+            // Add linked preset info if available
+            let presetItemType = this.subType === 'Stable-Diffusion' ? 'model' : (this.subType === 'LoRA' ? 'lora' : null);
```
mcmonkey4eva: `==` not `===`

src/wwwroot/js/genpage/gentab/models.js
```diff
@@ -894,8 +915,53 @@ function directSetModel(model) {
...
+        if (presetTitle) {
+            // Enforce mutual exclusivity - remove other models' presets
+            currentPresets = removeOtherModelPresets(model.name || modelName);
+            let presetData = allPresetsUnsorted?.find(p => p.title == presetTitle);
+            if (presetData) {
+                if (autoApply) {
+                    applyOnePreset(presetData, false); // false = not nested, allow LoRA preset triggering
+                    // After auto-applying, remove the preset from currentPresets so user can edit those params
+                    currentPresets = currentPresets.filter(p => p.title != presetData.title);
+                } else {
```
mcmonkey4eva: newline after `}`

src/wwwroot/js/genpage/gentab/presets.js
```diff
@@ -250,7 +289,17 @@ function selectInitialPresetList() {
     }
 }
 
-function applyOnePreset(preset) {
+function applyOnePreset(preset, isNestedPreset = false) {
+    // Collect LoRA updates before applying to handle merging
+    let presetLoras = preset.param_map.loras ? preset.param_map.loras.split(',').map(l => l.trim()).filter(l => l.length > 0) : [];
```
mcmonkey4eva: imo any changes to how loras are handled in presets is out-of-scope and should be a separate followup. This is already a very big PR, try to keep it focused to the core function.

src/Pages/_Generate/GenTabModals.cshtml
```diff
@@ -156,6 +156,19 @@
         </select></div>
         Description: <textarea class="auto-text auto-text-block translate" id="edit_model_description" rows="3" placeholder="Model description text..."></textarea>
         <div id="edit_model_image_block"></div>
+        <div id="edit_model_override_preset_block" class="input-group">
+            <div class="input-group-content flex-column">
+                <div class="d-flex flex-row">
```
mcmonkey4eva: `d-flex`/`flex-row` should not be used (Avoid any bootstrap classes not already explicitly in use)

src/WebAPI/AdminAPI.cs
```diff
@@ -55,6 +55,10 @@ public static JObject AutoConfigToParamData(AutoConfiguration config, bool hideR
         JObject output = [];
         foreach ((string key, AutoConfiguration.Internal.SingleFieldData data) in config.InternalData.SharedData.Fields)
         {
+            if (data.Field.GetCustomAttribute<SettingHiddenAttribute>() is not null)
```
mcmonkey4eva: why are there any code edits in this file? There should not be.

src/WebAPI/BasicAPIFeatures.cs
```diff
@@ -226,14 +226,29 @@ public static async Task<JObject> InstallConfirmWS(Session session, WebSocket so
...
         return new JObject()
         {
             ["user_name"] = session.User.UserID,
             ["presets"] = new JArray(session.User.GetAllPresets().Select(p => p.NetData()).ToArray()),
             ["language"] = session.User.Settings.Language,
             ["permissions"] = JArray.FromObject(session.User.GetPermissions()),
             ["starred_models"] = JObject.Parse(session.User.GetGenericData("starred_models", "full") ?? "{}"),
-            ["autocompletions"] = string.IsNullOrWhiteSpace(settings.Source) ? null : new JArray(AutoCompleteListHelper.GetData(settings.Source, settings.EscapeParens, settings.Suffix, settings.SpacingMode))
+            ["autocompletions"] = string.IsNullOrWhiteSpace(settings.Source) ? null : new JArray(AutoCompleteListHelper.GetData(settings.Source, settings.EscapeParens, settings.Suffix, settings.SpacingMode)),
+            ["itemPresetLinks"] = itemPresetLinks
```
mcmonkey4eva: basic format matching issue

src/WebAPI/BasicAPIFeatures.cs
```diff
@@ -226,14 +226,29 @@ public static async Task<JObject> InstallConfirmWS(Session session, WebSocket so
     public static async Task<JObject> GetMyUserData(Session session)
     {
         Settings.User.AutoCompleteData settings = session.User.Settings.AutoComplete;
+        JObject itemPresetLinks = new JObject();
+        try
+        {
+            string rawJson = session.User.GetGenericData("itempresetlinks", "data");
+            if (!string.IsNullOrWhiteSpace(rawJson))
+            {
+                itemPresetLinks = JObject.Parse(rawJson);
```
mcmonkey4eva: what does `item` mean here? Should probably be `model`?

src/WebAPI/ModelsAPI.cs
```diff
@@ -540,8 +542,8 @@ public static async Task<JObject> EditModelMetadata(Session session,
             actualModel.Metadata.LoraDefaultWeight = lora_default_weight;
             actualModel.Metadata.LoraDefaultConfinement = lora_default_confinement;
             actualModel.Metadata.PredictionType = string.IsNullOrWhiteSpace(prediction_type) ? null : prediction_type;
+            handler.ResetMetadataFrom(actualModel);
```
mcmonkey4eva: this is an actively dangerously line edit. Once again, _why are you editing code here at all?_

src/WebAPI/ModelsAPI.cs
```diff
@@ -889,4 +891,63 @@ void doMoveNow(string oldPath)
...
+            {
+                links.Remove(key);
+            }
+            string newJson = links.ToString(Newtonsoft.Json.Formatting.None);
+            session.User.SaveGenericData("itempresetlinks", "data", newJson);
+            return new JObject() { ["success"] = true, ["key"] = key, ["preset"] = presetTitle };
+        }
+        catch (Exception ex)
+        {
+            Logs.Error($"Error saving preset link: {ex.Message}");
```
mcmonkey4eva: `ReadableString()`

src/WebAPI/ModelsAPI.cs
```diff
@@ -889,4 +891,63 @@ void doMoveNow(string oldPath)
...
+            return new JObject() { ["success"] = false, ["error"] = ex.Message };
```
mcmonkey4eva: api routes intentionally do not return internal error data to the user like this

src/wwwroot/css/genpage.css
```diff
@@ -559,6 +559,10 @@ body {
     font-size: 80%;
     color: var(--text-soft);
 }
+.model_preset {
+    color: var(--text-soft);
+}
+
```
mcmonkey4eva: format

src/wwwroot/css/genpage.css
```diff
@@ -1660,3 +1664,8 @@ body {
     height: 1lh;
     padding-top: 0 !important;
 }
+#edit_model_override_preset button.disabled-dim {
```
mcmonkey4eva: seems oddly specific

src/wwwroot/js/genpage/gentab/models.js
```diff
@@ -894,8 +915,53 @@ function directSetModel(model) {
...
+        modelName = name;
     }
     reviseBackendFeatureSet();
+    
+    // Look up preset link using the full model name (getItemPresetLink handles extraction)
+    let presetTitle = getItemPresetLink('model', model.name || modelName);
+    try {
+        let autoApply = getUserSetting('ui.autoapplymodelpresets', false);
+        if (presetTitle) {
+            // Enforce mutual exclusivity - remove other models' presets
```
mcmonkey4eva: don't add line comments that just say what the line of code already says

src/wwwroot/js/genpage/gentab/models.js
```diff
@@ -1024,4 +1090,121 @@ function doModelInstallRequiredCheck() {
...
+ * 
+ * Examples:
+ * - 'SDXL/models/my_model.safetensors' -> 'my_model'
+ * - 'loras/my_lora.safetensors' -> 'my_lora'
+ * - 'my_model' -> 'my_model'
+ * 
+ * @param {string} itemName - The full model/LoRA name (path or filename)
+ * @returns {string} The base filename without path or extension
+ */
+function extractItemKeyNameFromPath(itemName) {
```
mcmonkey4eva: get rid of this, no. bad.

src/wwwroot/js/genpage/gentab/presets.js
```diff
@@ -10,6 +10,45 @@ class PresetHelpers {
     }
 }
 
+/** Manages preset links for models and LoRAs. User-scoped storage in User.UserUIData.ItemPresetLinks */
```
mcmonkey4eva: `User-scoped storage in User.UserUIData.ItemPresetLinks` this is just... strange text to put in a js class doc.

src/wwwroot/js/genpage/gentab/models.js
```diff
@@ -265,10 +267,17 @@ function save_edit_model() {
...
+    
     function complete() {
         genericRequest('EditModelMetadata', data, data => {
             curModelMenuBrowser.browser.lightRefresh();
         });
+        // Save preset link using centralized function
+        let presetSelect = document.getElementById('edit_model_override_preset_id');
+        let presetTitle = presetSelect?.value || '';
+        // Determine item type based on browser subType (LoRA, Embedding, etc) or model architecture
+        let itemType = curModelMenuBrowser.subType === 'LoRA' ? 'lora' : 'model';
```
mcmonkey4eva: What even - why? Just use the subType. These are well defined keys, why are you inventing new keys and interconverting? This is unneeded complexity.

src/Core/Settings.cs
```diff
@@ -391,6 +391,12 @@ public class UserUIData : AutoConfiguration
             [ConfigComment("If enabled, shifting to next/previous image (eg with arrow keys) in history or batch view,\ncycles at the ends (jumps from the start to the end or vice versa).\nIf disabled, shifting will simply stop at the ends.\nIf 'only arrow keys', cycling happens when you press the arrow keys, but not other actions (eg deleting an image will not cycle).")]
             [ManualSettingsOptions(Vals = ["true", "false", "only_arrows"], ManualNames = ["Enabled", "Disabled", "Only Arrow Keys"])]
             public string ImageShiftingCycles = "true";
+
+            [ConfigComment("When a model with a linked preset is selected, should the preset parameters be applied immediately? By default, the preset is just selected for review.")]
+            public bool AutoApplyModelPresets = false;
```
mcmonkey4eva: It'd probably make more sense for this to be a setting on each Preset rather than a global user setting. Some presets should autoapply, others shouldn't. The functionality would probably be beneficial for all presets rather than specific to model. This should aaalso probably be a separate PR.

src/WebAPI/BasicAPIFeatures.cs
richardtallent: corrected to snake case across the board

src/WebAPI/BasicAPIFeatures.cs
richardtallent: I was trying to make more clear that the ability to set a preset applies to both models and lora. Happy to rename if you have a better idea.

mcmonkey4eva: Base Models and LoRAs are both types of model

src/WebAPI/ModelsAPI.cs
richardtallent: My original idea was to use metadata. During that change, the ResetMetadataFrom() got inadvertently moved into the lock{} clause, and when I undid that change and switched to user settings, I missed that remaining change. I moved it back.

src/wwwroot/css/genpage.css
richardtallent: Yes, it was. Was trying to set a "not allowed" cursor but another rule setting the cursor had higher selectivity. Removed the ID selector and the cursor setting.

src/WebAPI/BasicAPIFeatures.cs
richardtallent: Fair enough. I was going by how the UI refers to them. Ran out of time tonight, will come back to this and remaining items as soon as I get a chance again.

src/wwwroot/js/genpage/gentab/models.js
richardtallent: What specifically are you concerned about here?

My concern was that as I reorganize my models, I don't want them to lose their presets just because they moved to a new folder path. That wasn't an issue when it was custom metadata, but as a user setting, need a durable key.

mcmonkey4eva: Maintaining model list vs user list sync is... an open question to be discussed, but the solution is definitely not this.

richardtallent: Removed. There is a similar cleanModelName function in the codebase now, but it was overridden with another function later in the same file. That version still at least removes the `.safetensors` extension, but that doesn't really move the needle on avoiding link loss, so I didn't call it either.

src/wwwroot/js/genpage/gentab/presets.js
richardtallent: Fair point, removed

src/Core/Settings.cs
richardtallent: Sounds good, removed in this form

richardtallent:
All comments addressed, please let me know if additional changes are needed.

PR #1118, opened 2025-11-07
Adds WASD image browse/star/delete functionality

jtreminio:
Adds WASD-based image browsing, star, and delete functionality.

* `W`: LEFT ARROW
* `S`: RIGHT ARROW
* `A`: STAR IMAGE
* `D`: DELETE IMAGE

Deleting the image requires pressing the `D` key twice within 500ms, and not performing any other action from this list in between key presses: `['mousedown', 'wheel', 'scroll', 'touchstart', 'pointerdown', 'contextmenu']`.

We can quickly change what keys map to what action. I realize A/D might be better for LEFT/RIGHT arrow functionality, but D (especially double D) maps to DELETE in vim and D is for delete so 🤷 .

Followup PRs (or maybe add to this one?):

* user-configurable keys - I imagine folks with non-qwerty keyboards might want something else, or someone without a left hand??
* user-configurable timeout - for folks who can't physically press a key twice within half a second of each other
* adding to documentation, and/or adding tooltip or info somewhere in GUI.

### DIFF
--- a/src/wwwroot/js/genpage/gentab/currentimagehandler.js
+++ b/src/wwwroot/js/genpage/gentab/currentimagehandler.js
@@ -533,20 +533,110 @@ function shiftToNextImagePreview(next = true, expand = false, isArrows = false) 
     return true;
 }
 
+let deleteKeyPrimed = false;
+let deleteKeyTimer = null;
+let deleteKeyCancelListeners = [];
+
+function cancelDeletePrime() {
+    if (!deleteKeyPrimed) {
+        return;
+    }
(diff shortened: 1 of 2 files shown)


mcmonkey4eva:
Delete/star hotkeys are a valid pending feature request this could address, W/S is ... quite odd. User settings to allow custom additional hotkeys would probably be the way. Maybe a dedicated "Hotkeys" sub-menu with some useful way to specify a list of keys (or hack it with a csv and fix it later I guess). Either way, that is a discussion to have as a feature request, not a code change.

PR #1120, opened 2025-11-08
Respect user VAE for Segment model

jtreminio:
Fixes bug where user setting for VAE is not being used for Segment model.


jtreminio:
@mcmonkey4eva would you consider adding constants for all these model IDs?

src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
```diff
@@ -1378,6 +1378,42 @@ JArray doMaskShrinkApply(WorkflowGenerator g, JArray imgIn)
                     (t2iModel, model, clip, vae) = g.CreateStandardModelLoader(t2iModel, "Refiner");
                     g.FinalLoadedModel = t2iModel;
                     g.FinalModel = model;
+                    string compat = t2iModel.ModelClass?.CompatClass;
```
mcmonkey4eva: this could shouldn't be here -- CreateStandardModelLoader should be grabbing the correct VAE

mcmonkey4eva:
This code still doesn't make sense. There's already code to load the correct VAE.

PR #1121, opened 2025-11-08
Highlight current image in the history container

jtreminio:
Also works when genning a new image and clicking the refresh history button.

https://github.com/user-attachments/assets/492389e8-af1a-445e-bc9d-7ee447bc2ab5


mcmonkey4eva:
Good improvement, seems to work in testing first try! Nice!

PR #1122, opened 2025-11-08
Fix selecting image on delete from end

jtreminio:
When `ImageShiftingCycles == Only Array Keys`, and deleting the first or last image, I expect the 2nd image from the current end to be selected. Instead, no image is selected and I must click another image again.

https://github.com/user-attachments/assets/641d624e-6eae-4cbe-b124-e833e922a16b

### DIFF
--- a/src/wwwroot/js/genpage/gentab/currentimagehandler.js
+++ b/src/wwwroot/js/genpage/gentab/currentimagehandler.js
@@ -478,15 +478,25 @@ function shiftToNextImagePreview(next = true, expand = false, isArrows = false) 
         let newIndex = index + (next ? 1 : -1);
         if (newIndex < 0) {
             if (!doCycle) {
-                return false;
+                newIndex = index + 1;
+                if (newIndex >= divs.length) {
+                    return false;
+                }
             }
-            newIndex = divs.length - 1;
+            else {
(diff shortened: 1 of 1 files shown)


jtreminio:
I get that you don’t personally like how this works, but as of last week the functionality feels broken compared to how it worked before my PR to maintain fullscreen modal open on delete.

My ask is a simple and basic use case: don’t deselect an image when deleting from the edge, don’t jump from one edge to the other when deleting. That’s how it worked before.

I do also understand that this code starts looking ugly, and it probably needs to live somewhere else, but that’s a result of the current JS functions that need slimming down and simplifying.

mcmonkey4eva:
A PR is not a feature request. Ref the contrib doc https://github.com/mcmonkeyprojects/SwarmUI/blob/master/CONTRIBUTING.md `You are expected [...] To have discussed your PR in advance (GitHub issue, or Discord) before making it`

We can talk about maybe adding another yet-more-options to that user setting, though I must emphasize this is an extremely minor behavioral change that you think kinda works nice and I think is an odd inconsistency of UX - this is not a hill to fight to the death on. At best it saves you one click with a very specific pattern of usage, at the cost of cluttering the settings menu and the code further.

PR #1158, opened 2025-12-01
Do frame interpolation only once at the end of all <extend>(Fixes #994)

### LINKED ISSUE #994: VFI happens before extend
### Expected Behavior

A video should get generated, then extended, then interpolated

### Actual Behavior

When using the video extend feature and frame interpolation combined. The generated workflow applies the frame interpolation to the first generated video, then uses that to extend.

So when generating 80 frames, then extending 80 frames with interpolation set to 2, the first half of the video is at 50% speed, the second half is normal speed (base model framerate, not interpolated)

### Steps to Reproduce

1. Load a model for text to video or image to video at for example 80 frames
2. Enable video extend and add <extend:80> in the prompt (or any other valid number of frames)
3. Enable frame interpolation for your video (in image to video or text to video section depending on what's used)
4. Generate, the result will have the first section interpolated, and the second section raw, the output uses the framerate of an uninterpolated video

### Debug Logs

-

### Other

Currently video frame interpolation is a separate setting in 2 parts of the UI, one for t2v and one for i2v. It might make more sense to have one setting, for example in advanced video (although that section is only visible with display advanced options). With the interpolation applied in the end instead of between video gen and video extend.

willhsmit:
* Add code to the I2V flow to skip frame interpolation if there are additional <extend> prompts (T2V already has this check).
* Add code to the extend flow to do frame interpolation after the last extend. There are separate T2V and I2V interpolation method and multiplier, so check whether VideoModel is set to pick the appropriate one.

Confirmed that for all of:
T2V with no extends and with 1-2 extends
I2V with no extends and with 1-2 extends

there is exactly one RIFE VFI node before the last SwarmSaveAnimationWS with the right multiplier, and the unmultiplied fps on any prior save and the multiplied fps on the last save.

Clipboard01

### DIFF
--- a/src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
+++ b/src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
@@ -1731,8 +1731,8 @@ public class WorkflowGeneratorSteps
                 };
                 g.CreateImageToVideo(genInfo);
                 videoFps = genInfo.VideoFPS;
-                bool willHaveFollowupVideo = prompt.Contains("<extend:");
-                if (!willHaveFollowupVideo && g.UserInput.TryGet(ComfyUIBackendExtension.VideoFrameInterpolationMethod, out string method) && g.UserInput.TryGet(ComfyUIBackendExtension.VideoFrameInterpolationMultiplier, out int mult) && mult > 1)
+                bool hasExtend = prompt.Contains("<extend:");
+                if (!hasExtend && g.UserInput.TryGet(ComfyUIBackendExtension.VideoFrameInterpolationMethod, out string method) && g.UserInput.TryGet(ComfyUIBackendExtension.VideoFrameInterpolationMultiplier, out int mult) && mult > 1)
                 {
                     if (g.UserInput.Get(T2IParamTypes.OutputIntermediateImages, false))
                     {
@@ -1758,7 +1758,7 @@ public class WorkflowGeneratorSteps
                     g.FinalImageOut = [bounced, 0];
                 }
                 string nodeId = "9";
-                if (prompt.Contains("<extend:"))
+                if (hasExtend)
                 {
                     nodeId = $"{g.GetStableDynamicID(50000, 0)}";
                 }


src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
```diff
@@ -1371,6 +1371,33 @@ public JArray DoInterpolation(JArray imageIn, string method, double mult)
         return imageIn;
     }
 
+    /// <summary>Get the T2V or I2V frame interpolation method as appropriate.</summary>
+    public bool GetCurrentVideoFrameInterpolationMethod(out string method)
+    {
+        if (UserInput.TryGet(T2IParamTypes.VideoModel, out _))
+        {
+            return UserInput.TryGet(ComfyUIBackendExtension.VideoFrameInterpolationMethod, out method);
```
mcmonkey4eva: this function ain't right.

src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
```diff
@@ -1731,7 +1731,11 @@ void altLatent(WorkflowGenerator.ImageToVideoGenInfo genInfo)
                 };
                 g.CreateImageToVideo(genInfo);
                 videoFps = genInfo.VideoFPS;
-                if (g.UserInput.TryGet(ComfyUIBackendExtension.VideoFrameInterpolationMethod, out string method) && g.UserInput.TryGet(ComfyUIBackendExtension.VideoFrameInterpolationMultiplier, out int mult) && mult > 1)
+                bool willHaveFollowupVideo = g.UserInput.Get(T2IParamTypes.Prompt, "").Contains("<extend:");
+                if (!willHaveFollowupVideo
+                    && g.UserInput.TryGet(ComfyUIBackendExtension.VideoFrameInterpolationMethod, out string method)
+                    && g.UserInput.TryGet(ComfyUIBackendExtension.VideoFrameInterpolationMultiplier, out int mult)
+                    && mult > 1)
```
mcmonkey4eva: this should not be reformatted like this

src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
```diff
@@ -1877,6 +1881,12 @@ void altLatent(WorkflowGenerator.ImageToVideoGenInfo genInfo)
                     conjoinedLast = [batchedNode, 0];
                 }
                 g.FinalImageOut = conjoinedLast;
+                if (g.GetCurrentVideoFrameInterpolationMethod(out string method) && g.GetCurrentVideoFrameInterpolationMultiplier(out int mult) && mult > 1)
+                {
+                    // Do we need an intermediate save here, or are the saves at each extend enough?
```
mcmonkey4eva: yeah a save-intermediate before interp would be wanted

willhsmit: Added the save.

src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
willhsmit: Restored one-line format.

src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
willhsmit: Not sure I follow. Is your comment:
* Behavior should be similar, but inlined at the callsite
* Behavior should be similar, but organized differently here, like a ternary on the parameter name
* Code should be using something different than existence of a VideoModel to determine whether to use I2V or T2V interpolation settings when extend is present
* Code should always choose the same parameter (I2V?) to govern interpolation when extends are present, regardless of I2V or T2V on the initial gen.

src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
```diff
@@ -1731,7 +1731,8 @@ void altLatent(WorkflowGenerator.ImageToVideoGenInfo genInfo)
                 };
                 g.CreateImageToVideo(genInfo);
                 videoFps = genInfo.VideoFPS;
-                if (g.UserInput.TryGet(ComfyUIBackendExtension.VideoFrameInterpolationMethod, out string method) && g.UserInput.TryGet(ComfyUIBackendExtension.VideoFrameInterpolationMultiplier, out int mult) && mult > 1)
+                bool willHaveFollowupVideo = g.UserInput.Get(T2IParamTypes.Prompt, "").Contains("<extend:");
```
mcmonkey4eva: this should match line 1761

src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
mcmonkey4eva: In the current format this is a strange hack.

Per my comment on the issue post, arguably probably these params should just be unified to a single parameter pair under Advanced Video rather than split between t2v and i2v, which would cure this issue.

willhsmit: Removed Text2VideoFrameInterpolation.*, moved VideoFrameInterpolation.* (and the install button) into Advanced Video, and changed all the calls to VideoFrameInterpolation.*. Tested the i2v/t2v extend/noextend flows and delete/reinstall of the ComfyUI-Frame-Interpolation and it seems okay.

That change affects a couple UI files but WorkflowGenerator.cs no longer needs to be in the PR.

src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
willhsmit: Done.

PR #1162, opened 2025-12-02
Initial Ovis support

jtreminio:
CleanShot 2025-12-02 at 11 16 47

### DIFF
--- a/docs/Model Support.md
+++ b/docs/Model Support.md
@@ -30,7 +30,7 @@ Old or bad options also tracked listed:
 [Nvidia Sana](#nvidia-sana) | DiT | 2024 | NVIDIA | 1.6B | No | Just Bad |
 [Nvidia Cosmos Predict2](#cosmos-predict2) | DiT | 2025 | NVIDIA | 2B/14B | Partial | Just Bad |
 [HiDream i1](#hidream-i1) | MMDiT | 2025 | HiDream AI (Vivago) | 17B | Minimal | Good Quality, lost community attention |
-[Ovis](#ovis) | MMDiT | 2025 | AIDC-AI (Alibaba) | 7B | No | Image quality at slightly more than base SDXL levels, supposedly good (?) at text |
+[Ovis](#ovis) | MMDiT | 2025 | AIDC-AI (Alibaba) | 7B | No | Passable quality, but outclassed on launch |
 
 - **Architecture** is the fundamental machine learning structure used for the model, UNet's were used in the past but DiT (Diffusion Transformers) are the modern choice
 - **Scale** is how big the model is - "B" for "Billion", so for example "2B" means "Two billion parameters".
@@ -608,10 +608,10 @@ These steps are not friendly to beginners (if Sana gains popularity, likely more
 - Uses the Flux.1 VAE
 - **Parameters:**
     - **Prompt:** Supports general prompting in any format just fine. Speaks English and Chinese.
-    - **Sampler:** `Euler`
-    - **Scheduler:** `Beta`
-    - **CFG Scale:** `5`
-    - **Steps:** Recommended `50`, seems "fine" at `25`
+    - **Sampler:** Default is fine (`Euler`)
+    - **Scheduler:** Default works, but `Beta` may be better
+    - **CFG Scale:** Normal CFG ranges, `5` is the official recommendation
+    - **Steps:** Normal step counts (eg `20`), but they recommend `50`
     - **Resolution:** Side length `1024`. Quickly breaks above that.
 
 # Video Models
--- a/src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorModelSupport.cs
+++ b/src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorModelSupport.cs
@@ -467,7 +467,7 @@ public partial class WorkflowGenerator
 
         public string GetOvisQwenModel()
         {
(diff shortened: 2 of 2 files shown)


src/Text2Image/T2IModelClassSorter.cs
```diff
@@ -525,6 +527,12 @@ JToken GetEmbeddingKey(JObject h)
         {
             return isZImageControlNet(h);
         }});
+        // ====================== Ovis ======================
+        Register(new() { ID = "ovis", CompatClass = CompatOvis, Name = "Ovis", StandardWidth = 1024, StandardHeight = 1024, IsThisModelOfClass = (m, h) =>
+        {
+            return isLumina2(h) && isOvis(h);
+        }});
+        Logs.Info("Registered Ovis model class.");
```
mcmonkey4eva: stray debug

src/Text2Image/T2IModelClassSorter.cs
```diff
@@ -525,6 +527,12 @@ JToken GetEmbeddingKey(JObject h)
         {
             return isZImageControlNet(h);
         }});
+        // ====================== Ovis ======================
```
mcmonkey4eva: I think this can go under the random dump, per news-talk discussion it doesn't seem likely to be a noteworthy model

src/Text2Image/T2IModelClassSorter.cs
```diff
@@ -525,6 +527,12 @@ JToken GetEmbeddingKey(JObject h)
...
+        Register(new() { ID = "ovis", CompatClass = CompatOvis, Name = "Ovis", StandardWidth = 1024, StandardHeight = 1024, IsThisModelOfClass = (m, h) =>
+        {
+            return isLumina2(h) && isOvis(h);
```
mcmonkey4eva: if `isLumina2` matches, make sure to also `&& !isOvis` on the other lumina (and make sure it's even correctly key-separated from base lumina and z-image)

jtreminio:
```
$ venv/bin/python - << 'EOF'
> import json, struct, os
>
> paths = {
>     "ovis": os.path.expanduser("~/data/AIConfigs/Models/Stable-Diffusion/ovis_image_bf16.safetensors"),
>     "zimage": os.path.expanduser("~/data/AIConfigs/Models/diffusion_models/z-image/z_image_turbo_bf16.safetensors"),
> }
>
> keys_to_check = [
>     "context_refiner.0.attention.k_norm.weight",
>     "double_blocks.0.img_mlp.gate_proj.weight",
>     "txt_norm.scale",
> ]
>
> def header_keys(path):
>     with open(path, "rb") as f:
>         header_len = struct.unpack("<Q", f.read(8))[0]
>         header_json = f.read(header_len).decode("utf-8")
>         header = json.loads(header_json)
>         return set(header.keys())
>
> for name, path in paths.items():
>     print(f"FILE {name}: {path}")
>     ks = header_keys(path)
>     for k in keys_to_check:
>         print(f"  {k}: {k in ks}")
>     print()
> EOF

FILE ovis: /home/jtreminio/data/AIConfigs/Models/Stable-Diffusion/ovis_image_bf16.safetensors
  context_refiner.0.attention.k_norm.weight: False
  double_blocks.0.img_mlp.gate_proj.weight: True
  txt_norm.scale: True

FILE zimage: /home/jtreminio/data/AIConfigs/Models/diffusion_models/z-image/z_image_turbo_bf16.safetensors
  context_refiner.0.attention.k_norm.weight: True
  double_blocks.0.img_mlp.gate_proj.weight: False
  txt_norm.scale: False
  ```

mcmonkey4eva: You're missing base Lumina2 from there - https://huggingface.co/Comfy-Org/Lumina_Image_2.0_Repackaged/blob/main/all_in_one/lumina_2.safetensors

... which makes me realize, the zimage check is bad, base lumina2 has `model.diffusion_model.context_refiner.0.attention.k_norm.weight`
In fact, keys look about identical. So probably need shape checks, cap_embedder and that context refiner k_norm have clearly different shapes

mcmonkey4eva: actually, Ovis https://huggingface.co/Comfy-Org/Ovis-Image/blob/main/split_files/diffusion_models/ovis_image_bf16.safetensors doesn't even have the lumina2 matcher key, so that detection doesn't make sense at all

mcmonkey4eva:
I'd guess this also needs to be added to the sigma shift support next to zimage

src/Text2Image/T2IModelClassSorter.cs
jtreminio: The ovis match comes from the [ComfyUI PR](https://github.com/comfyanonymous/ComfyUI/commit/878db3a727c1c6049bc1c4959cdfabc35eaf3d56#diff-cf59676997a5af41382b3992a08c35bd35f941aa5ea55b4e17e723ccc78d0932R276-R279)

mcmonkey4eva: `return isLumina2(h) && isOvis(h);` this line cannot work, _because of the lumina2 part of it_

mcmonkey4eva: hit utilities -> reset all metadata and watch logs, should tell you the model failed to match anything (or detect as something else potentially)

jtreminio: I've got several VAE and text encoders that show up, and a few Illustrious models, but not Ovis or zit.

PR #1173, opened 2025-12-06
Add option to hide metadata in fullscreen by default

jtreminio:
Adds user option with simple objective: hide metadata when opening an image in fullscreen. All other behavior remains: zoom out shows metadata, left/right keeps current flag, closing/reopening resets to default.

https://github.com/user-attachments/assets/1cdb3a58-d67b-4295-aa2f-3e4732dcbe00

### DIFF
--- a/src/wwwroot/js/genpage/gentab/currentimagehandler.js
+++ b/src/wwwroot/js/genpage/gentab/currentimagehandler.js
@@ -180,10 +180,10 @@ class ImageFullViewHelper {
         else {
             img.style.imageRendering = '';
         }
-        if (newHeight > 101 && this.showMetadata) {
+        if (newHeight > 100.1) {
             this.toggleMetadataVisibility(false);
         }
-        else if (newHeight < 101 && !this.showMetadata && zoom < 1) {
+        else if (newHeight < 100.1) {
             this.toggleMetadataVisibility(true);
         }
         container.style.cursor = 'grab';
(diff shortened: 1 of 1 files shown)


src/wwwroot/js/genpage/gentab/currentimagehandler.js
```diff
@@ -180,10 +180,10 @@ class ImageFullViewHelper {
         else {
             img.style.imageRendering = '';
         }
-        if (newHeight > 101) {
+        if (newHeight > 101 && this.showMetadata) {
             this.toggleMetadataVisibility(false);
         }
-        else if (newHeight < 101) {
+        else if (newHeight < 101 && !this.showMetadata && zoom < 1) {
```
jtreminio: This change is required. May seem weird, but add a `console.log()` to both cases to verify.

We only want to trigger `this.toggleMetadataVisibility()` when necessary - meaning:

* if an image shows the metadata, zooming out should not trigger toggleMetadataVisibility
* if an image does not show the metadata, zooming out should trigger toggleMetadataVisibility when zoom level < 101
* if an image shows the metadata, zooming in should trigger toggleMetadataVisibility when zoom level > 101
* if an image does not show the metadata, zooming in should not trigger toggleMetadataVisibility

mcmonkey4eva: ... Why? Retriggering it has no side effects.

jtreminio: It does if the setting is enabled to hide metadata by default, and you zoom in right after opening the image. The metadata will flash before being hidden again.

This is the cleanest way to avoid having to keep track of further state.

mcmonkey4eva: ohh no the impl here is wrong.
You need to set the zoom to a different value according to user preference, not just default visible/hidden.

PR #1175, opened 2025-12-09
UI Support for the SeedVariance node

rlerdorf:
# Add SeedVarianceEnhancer Integration

Adds UI support for the https://github.com/ChangeTheConstants/SeedVarianceEnhancer ComfyUI custom node, which improves seed variance on models with low natural variance (like Z-Image Turbo).

## What it does

SeedVarianceEnhancer adds controlled noise to conditioning embeddings, resulting in more diverse outputs when changing seeds. This is particularly useful for turbo/distilled models that tend to produce very similar images across different seeds.

| Parameter                  | Description                                                                             | Default |
|----------------------------|-----------------------------------------------------------------------------------------|---------|
| Seed Variance Strength     | Noise scale added to embeddings. Higher = more variance.                                | 20      |
| Seed Variance Randomize %  | Percentage of embedding values to modify. Lower = more stable anatomy.                  | 50%     |
| Seed Variance Switchover % | When to stop using noisy embeddings (as % of steps). Lower = noise only on early steps. | 20%     |
| Seed Variance Seed         | Seed for noise generation. -1 = random.                                                 | -1      |

## Usage Notes
- Parameters appear in Advanced Sampling (requires "Advanced" toggle)
- Only visible when the SeedVarianceEnhancer node is installed in ComfyUI
- Applied to positive conditioning only
- For LoRAs with anatomy issues, try: Strength 10-15, Randomize 20-30%, Switchover 10-15%

### DIFF
--- a/src/BuiltinExtensions/ComfyUIBackend/ComfyUIBackendExtension.cs
+++ b/src/BuiltinExtensions/ComfyUIBackend/ComfyUIBackendExtension.cs
@@ -68,7 +68,8 @@ public class ComfyUIBackendExtension : Extension
         ["TeaCache"] = "teacache",
         ["TeaCacheForVidGen"] = "teacache",
         ["TeaCacheForImgGen"] = "teacache_oldvers",
-        ["OverrideCLIPDevice"] = "set_clip_device"
+        ["OverrideCLIPDevice"] = "set_clip_device",
+        ["SeedVarianceEnhancer"] = "seed_variance_enhancer"
     };
 
     /// <inheritdoc/>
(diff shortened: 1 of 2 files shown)


mcmonkey4eva:
That node has a funny dirty trick of noising the cond embeds.

In base swarm theres another trick supported already <https://github.com/mcmonkeyprojects/SwarmUI/blob/master/docs/Model%20Support.md#z-image-turbo-seed-variety-trick>

Since this is a custom node, it should probably be an extension rather than core

Issue #700, opened 2025-04-09
Add a star system for the Presets


lerestesunira:
### Feature Idea

I have a lot of presets, it will be nice to have a star system for this categorie too! 
(When you are on the Presets tab, another cool thing will be to have the possibilities to create a folder on the left side, where there is the root list tree, and drag and drop the presets in it).

hansneddyanto:
Taking this

XDcsy:
@hansneddyanto Hi, I just noticed this issue has been claimed by you. I recently worked on this and opened a PR. Happy to adjust if it overlaps your work ：)

hansneddyanto:
Hi @XDcsy I'm currently still working on it, since I've been quite busy with personal things this past month. I've looked at your PR, and there's definitely quite a few things noteworthy that I can adjust into my code. I'll combine our code together (since there are quite a few overlaps) and open the PR after the holidays.

On another note, is there any way I can contact you so that it's easier to discuss on this? Happy holidays!! :D

mcmonkey4eva:
You should be in the swarm discord https://discord.gg/q2y38cqjNw to discuss any development

XDcsy:
@hansneddyanto  I'm glad it's helpful. I'm closing the PR and feel free to merge any parts of it that is useful into yours.
You can find me by `@sy3003` in the Discord server. Happy holidays!

mcmonkey4eva:
Fixed+Merged @hansneddyanto 's PR, you can star presets now

PR #1196, opened 2025-12-25
add star system for presets

XDcsy:
Implements #700, following the existing logic and UI patterns of the model star feature to ensure consistency.

BTW, I noticed that the current model star feature does not preserve the starred status after a model is renamed. In this PR, the starred status is preserved when a preset is renamed.

I did not modify the model-related behavior to keep this PR focused, but I would be happy to submit another one if needed.

Has been manually tested on Linux.


mcmonkey4eva:
This is the wrong approach -- Presets are already user-stored data, so they don't need a separate storage the way models do. They just need a boolean on the preset data

hansneddyanto:
Hi @mcmonkey4eva Just wanted to clarify, the approach should be having an additional boolean column in t2i_presets (something like "IsStarred") to indicate whether a preset is starred or not ya instead of storing it in GenericData?

mcmonkey4eva:
not a "column", a "field", not in "t2i_presets" in `T2IPreset`, but yes.

PR #1203, opened 2025-12-29
Enlarge Plus Sign Gen Button

CreativeAlloy:
Edited some of the `genpage.css` code to make the plus sign button to the left of the prompt box look bigger and more noticeable. Added margin-right to the textbox itself to ensure it doesn't go behind the Generate Button.

Before & After attached.
Before:
image
After:
image


CreativeAlloy:
For whatever reason this also included `bg.json` which is from an incomplete Bulgarian localization for Swarm. If it's no issue, it can be included, but let me know if I should close this PR.

PR #1210, opened 2026-01-02
Fixes zoom in shows metadata on widescreen images

jtreminio:
Currently if you open a widescreen image in fullscreen modal, zooming in will display the metadata and zoom _out_ the image. User must then zoom in again to get past the 100% threshold to hide the metadata.

This fix works to only show the metadata when zooming OUT and image is < 1.


mcmonkey4eva:
valid issue (enabled `DefaultHideMetadataInFullview` with wide images wonks the format), bad solution (breaking the stability of metadata visibility). Committed a more stable fix instead (detect wide images and don't hide metadata if it doesn't make sense to hide)

PR #1238, opened 2026-01-13
Truncates wildcard data in /API/ListModels

jtreminio:
On a system with > 34GB of wildcard data, setting depth to maximum now responds in under 2 seconds, vs previous crashing behavior.

### DIFF
--- a/src/wwwroot/js/genpage/gentab/wildcards.js
+++ b/src/wwwroot/js/genpage/gentab/wildcards.js
@@ -317,7 +317,14 @@ class WildcardHelpers {
             delete this.wildcardDataCache[name + "____READ_NOW"];
         }
         genericRequest('DescribeModel', { subtype: 'Wildcards', modelName: name }, data => {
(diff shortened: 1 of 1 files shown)


src/Utils/WildcardsHelper.cs
```diff
@@ -80,19 +79,33 @@ public static void Refresh()
         }
     }
 
+    public const int ListMaxChars = 512;
+
     /// <summary>Gets the wildcard data for the specified exact wildcard name.</summary>
-    public static Wildcard GetWildcard(string name)
+    public static Wildcard GetWildcard(string name, bool truncate = false)
```
mcmonkey4eva: truncate selection should be in `GetNetObject` not here

mcmonkey4eva:
This PR will misbehave on most of the content of wildcards.js when triggered.
notably the duplicate and test functions will be wrong. Possibly other things. Please search for usages of removed variables.

jtreminio:
Confirmed tests pass. Sorry about that, I ended up looking for an automated test suite and then got sidetracked and didn't come back to them!

PR #1239, opened 2026-01-16
Adds support for LTX-2 image-to-video with upscale

jtreminio:
This adds support for image-to-video using LTX-2, and the upscaler.

[ltx2-image-to-video-with-upscale.json](https://github.com/user-attachments/files/24657142/ltx2-image-to-video-with-upscale.json)

The last part of the workflow:

CleanShot 2026-01-15 at 19 16 38

CleanShot 2026-01-15 at 19 18 46

Had to add some checks to avoid double-upscaling (I ran into that myself, ended up trying to gen a video 4x the original image size). I believe I've followed your coding standards to the letter.


mcmonkey4eva:
This seems like a lot of special casing, even fully reinterpreting parameters depending on context. This needs more work on the planning side rather than the coding side.

The existing plan is to remove i2v as a whole separated special case, and instead make it so selecting an i2v model as base model you can just i2v as easily as you t2v, and thus you can also refiner as normal.

Issue #1119, opened 2025-11-07
Not adding "Text Match" in Auto-Segment And Refine generates broken syntax


jtreminio:
### Expected Behavior

If you leave "Text Match" empty in the Segment popup what's inserted is:

```<segment:,0.6,0.5>```

I didn't catch this at first, and when I tried to view the ComfyUI workflow it generated a broken thing:

<img width="547" height="684" alt="Image" src="https://github.com/user-attachments/assets/71b4348e-d1f5-49b1-8ed6-fa196e041aea" />

Will want to add validation:

1) prevent clicking "add" button unless required fields are populated, or
2) do not add to prompt unless required fields are populated, or
3) have a default value in the "Text Match" text field

Additionally, add something to the UI that flags a field as required (Bootstrap has something, or can just use `*` character)

### Actual Behavior

When "Text Match" is left blank, ComfyUI generates a broken workflow.

### Steps to Reproduce

Click "Auto Segment Refinement", then click "add" button. See broken prompt:

```<segment:,0.6,0.5>```

### Debug Logs

None necessary

mcmonkey4eva:
I think this needs two fixes:
- Disable the 'Add' button when there's not valid inputs
- Handle the bad segment input properly serverside (probably just let it emit the empty matching node)

mcmonkey4eva:
fixed by https://github.com/mcmonkeyprojects/SwarmUI/pull/1248

PR #1248, opened 2026-01-19
Fixed - Not adding "Text Match" in Auto-Segment And Refine generates broken syntax #1119

lucyunidev:
▲ Frontend
In prompttools.js
  - Added validation to prevent submitting empty "Text Match" field while using CLIP-Seg
  - Disables "Add" button when Text Match is empty (for CLIP-Seg only)

▲ Backend
In WorkflowGeneratorSteps.cs, when segment syntax has empty text match (e.g., `<segment:,0.6,0.5>`), creates an empty SwarmClipSeg node instead of breaking

▲  Testing
- I checked that empty Text Match field is blocked with proper error message and "Add" button is disabled when required field is empty
- I checked that server-side handles empty segment input without errors

### DIFF
--- a/src/wwwroot/js/genpage/gentab/prompttools.js
+++ b/src/wwwroot/js/genpage/gentab/prompttools.js
@@ -607,11 +607,6 @@ class PromptPlusButton {
         let modelText = this.segmentModalModelSelect.value;
         if (modelText == "CLIP-Seg") {
             modelText = this.segmentModalTextMatch.value.trim();
-            if (!modelText) {
-                showError(translate("Text Match is required when using CLIP-Seg. Please enter text to match against in the image."));
-                this.segmentModalTextMatch.focus();
-                return;
-            }
         }
         else { // YOLO
             if (parseInt(this.segmentModalYoloId.value) > 0) {


global.json
```diff
@@ -0,0 +1,7 @@
+{
```
mcmonkey4eva: what is this file? get rid of it

src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
```diff
@@ -1457,10 +1457,23 @@ void RunSegmentationProcessing(WorkflowGenerator g, bool isBeforeRefiner)
...
                     string segmentNode = null;
+                    if (segmentSections.Length == 0)
+                    {
+                        segmentSections = [""];
+                    }
                     foreach (string dataText in segmentSections)
                     {
                         string newSegmentNode = null;
-                        if (dataText.StartsWith("yolo-"))
+                        if (string.IsNullOrWhiteSpace(dataText))
```
mcmonkey4eva: this check does nothing and can be removed

src/wwwroot/js/genpage/gentab/prompttools.js
```diff
@@ -582,18 +584,76 @@ class PromptPlusButton {
...
+            }
+            this.updateSegmentModalAddButton();
+        }
+    }
+
+    updateSegmentModalAddButton() {
+        let modal = document.getElementById('text_prompt_segment_modal');
+        if (!modal) return;
+        let footer = modal.querySelector('.modal-footer');
+        if (!footer) return;
```
mcmonkey4eva: these checks make no sense (why would the elems not exist?), and the relevant elements are stored in fields already

src/wwwroot/js/genpage/gentab/prompttools.js
```diff
@@ -512,6 +513,7 @@ class PromptPlusButton {
             this.segmentModalClear();
             this.segmentModalProcessChanges();
             $('#text_prompt_segment_modal').modal('show');
+            setTimeout(() => this.updateSegmentModalAddButton(), 100);
```
mcmonkey4eva: why is this a timeout?

src/wwwroot/js/genpage/gentab/prompttools.js
```diff
@@ -582,18 +584,76 @@ class PromptPlusButton {
...
             this.segmentModalTextMatch.placeholder = text;
             this.segmentModalTextMatch.title = text;
+            this.segmentModalTextMatch.required = true;
+            let label = findParentOfClass(this.segmentModalTextMatch, 'auto-input').querySelector('.auto-input-name');
+            if (label) {
+                let existingAsterisk = label.querySelector('.required-asterisk');
+                if (!existingAsterisk) {
+                    let popoverButton = label.querySelector('.auto-input-qbutton, .info-popover-button');
+                    let asterisk = document.createElement('span');
+                    asterisk.className = 'required-asterisk';
```
mcmonkey4eva: i get what you're going for here but doesn't really fit the UI design, get rid of the asterisk thing

src/wwwroot/js/genpage/gentab/prompttools.js
```diff
@@ -582,18 +584,76 @@ class PromptPlusButton {
...
+        let footer = modal.querySelector('.modal-footer');
+        if (!footer) return;
+        let addButton = footer.querySelector('.btn-primary');
+        if (!addButton || addButton.textContent.includes('Cancel')) return;
+        
+        if (this.segmentModalModelSelect && this.segmentModalModelSelect.value == 'CLIP-Seg') {
+            let textMatch = this.segmentModalTextMatch ? this.segmentModalTextMatch.value.trim() : '';
+            if (!textMatch) {
+                addButton.disabled = true;
+                addButton.setAttribute('title', translate("Text Match is required when using CLIP-Seg"));
```
mcmonkey4eva: the standard modal design in swarmui is to have a little red text section next to the buttons, copy it from any other modal that has it in use, eg the preset import modal

global.json
lucyDev5: i use windows 11 and by default dotnet version 9 installed.
so i installed dotnet 8 and make this use version 8.

mcmonkey4eva: dotnet version is defined in the csproj, not a json file

src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
lucyDev5: good catch! i will removed the redundant check.

src/wwwroot/js/genpage/gentab/prompttools.js
lucyDev5: okay. i will simplify to directly query for the button.

src/wwwroot/js/genpage/gentab/prompttools.js
lucyDev5: got it.

src/wwwroot/js/genpage/gentab/prompttools.js
lucyDev5: i added this as a defensive measure, assuming the modal needed to be fully shown before updating the button state.
but that was unnecessary - the elements are already in the DOM, i will remove it.

src/wwwroot/js/genpage/gentab/prompttools.js
lucyDev5: got it.

src/wwwroot/js/genpage/gentab/prompttools.js
```diff
@@ -582,18 +584,47 @@ class PromptPlusButton {
...
             findParentOfClass(this.segmentModalTextMatch, 'auto-input').style.display = 'none';
             findParentOfClass(this.segmentModalYoloId, 'auto-input').style.display = '';
             findParentOfClass(this.segmentModalClassIds, 'auto-input').style.display = '';
+            this.segmentModalTextMatch.required = false;
+            this.updateSegmentModalAddButton();
+        }
+    }
+
+    updateSegmentModalAddButton() {
+        let addButton = document.querySelector('#text_prompt_segment_modal .modal-footer .btn-primary');
```
mcmonkey4eva: put these elements as a field like the rest of the elems

src/wwwroot/js/genpage/gentab/prompttools.js
```diff
@@ -582,18 +584,47 @@ class PromptPlusButton {
...
+        if (!addButton || addButton.textContent.includes('Cancel')) return;
```
mcmonkey4eva: this check makes no sense

src/wwwroot/js/genpage/gentab/prompttools.js
```diff
@@ -582,18 +584,47 @@ class PromptPlusButton {
...
+        let errorBox = getRequiredElementById('text_prompt_segment_modal_error');
+        
+        if (this.segmentModalModelSelect && this.segmentModalModelSelect.value == 'CLIP-Seg') {
```
mcmonkey4eva: stop checking if elements exist, _they do exist_

src/wwwroot/js/genpage/gentab/prompttools.js
```diff
@@ -582,18 +584,47 @@ class PromptPlusButton {
...
+            let textMatch = this.segmentModalTextMatch ? this.segmentModalTextMatch.value.trim() : '';
+            if (!textMatch) {
+                addButton.disabled = true;
+                errorBox.innerText = translate("Text Match is required when using CLIP-Seg");
+            } else {
```
mcmonkey4eva: format

src/wwwroot/js/genpage/gentab/prompttools.js
```diff
@@ -582,18 +584,47 @@ class PromptPlusButton {
             let text = translate("Text to match against in the image");
             this.segmentModalTextMatch.placeholder = text;
             this.segmentModalTextMatch.title = text;
+            this.segmentModalTextMatch.required = true;
```
mcmonkey4eva: not sure toggling this does anything relevant

src/wwwroot/js/genpage/gentab/prompttools.js
```diff
@@ -444,12 +444,15 @@ class PromptPlusButton {
...
         this.segmentModalModelSelect.addEventListener('change', () => this.segmentModalProcessChanges());
         this.segmentModalTextMatch = getRequiredElementById('text_prompt_segment_textmatch');
+        this.segmentModalTextMatch.addEventListener('input', () => this.updateSegmentModalAddButton());
         this.segmentModalClassIds = getRequiredElementById('text_prompt_segment_classids');
         this.segmentModalYoloId = getRequiredElementById('text_prompt_segment_yoloid');
         this.segmentModalCreativity = getRequiredElementById('text_prompt_segment_creativity');
         this.segmentModalThreshold = getRequiredElementById('text_prompt_segment_threshold');
         this.segmentModalInvertMask = getRequiredElementById('text_prompt_segment_invert_mask');
         this.segmentModalMainText = getRequiredElementById('text_prompt_segment_gentext');
+        this.segmentModalAddButton = document.querySelector('#text_prompt_segment_modal .modal-footer .btn-primary');
```
mcmonkey4eva: use getrequiredelementbyid

src/wwwroot/js/genpage/gentab/prompttools.js
```diff
@@ -582,18 +586,43 @@ class PromptPlusButton {
...
+    }
+
+    updateSegmentModalAddButton() {
+        if (this.segmentModalModelSelect.value == 'CLIP-Seg') {
+            let textMatch = this.segmentModalTextMatch.value.trim();
+            if (!textMatch) {
+                this.segmentModalAddButton.disabled = true;
+                this.segmentModalErrorBox.innerText = translate("Text Match is required when using CLIP-Seg");
+            }
+            else {
```
mcmonkey4eva: the two ifs can be combined into one line and not need duplicate else blocks

PR #1256, opened 2026-01-26
Make LoRA trigger phrases selectable. 

Reaper176:
This makes each word in the coma separated lists selectable  individually. Once you have selected each word you would like to copy you can click the button at the end of the list and it will copy only those selected words/phrases to the clipboard. If you want to copy the entire thing just click the copy button and it will copy everything.

I do a lot of jank stuff so maybe someone can clean this and make it nicer?

### DIFF
--- a/src/wwwroot/css/genpage.css
+++ b/src/wwwroot/css/genpage.css
@@ -1708,3 +1708,38 @@ body {
     height: 1lh;
     padding-top: 0 !important;
 }
+
+.trigger-phrase-item {
+    display: inline-block;
+    margin-right: 0.3rem;
+    padding: 0.15rem 0.35rem;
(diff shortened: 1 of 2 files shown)


mcmonkey4eva:
I don't think this is a friendly UI for the majority case, maybe as a special option or something?

Golden rule of PRs, discuss ideas _before_ writing code.

Reaper176:
> I don't think this is a friendly UI for the majority case, maybe as a special option or something?
> 
> Golden rule of PRs, discuss ideas _before_ writing code.

Sorry, I wrote these changes for my self and simply wanted to offer them to other people since this option is available in comfy, auto1111, reforge, forge classic, sd.next and invoke. Again, my apologies.

PR #1264, opened 2026-01-30
More robust image order enforcement

jtreminio:
Fixes a UI ordering race (introduced by my #1261) in the batch thumbnail strip when queueing multiple generations.

We now use `request_id` to keep reordering scoped to a single generation and avoid moving a just-finished image ahead of anything still generating.

Tracking is done via a simple `data-is_generating` flag, vs continuously checking DOM CSS class.

### DIFF
--- a/src/wwwroot/js/genpage/helpers/generatehandler.js
+++ b/src/wwwroot/js/genpage/helpers/generatehandler.js
@@ -210,7 +210,8 @@ class GenerateHandler {
                         if (insertBefore && insertBefore != div) {
                             div.parentElement.insertBefore(div, insertBefore);
                         }
-                    } else if (div.parentElement.firstElementChild != div) {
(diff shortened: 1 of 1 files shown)


src/wwwroot/js/genpage/helpers/generatehandler.js
```diff
@@ -189,12 +189,30 @@ class GenerateHandler {
...
+                            }
+                            if (c.dataset.request_id == data.request_id) {
+                                insertBefore = c;
+                                break;
+                            }
+                        }
+                        if (insertBefore && insertBefore != div) {
+                            div.parentElement.insertBefore(div, insertBefore);
+                        }
+                    } else if (div.parentElement.firstElementChild != div) {
```
mcmonkey4eva: format

PR #1271, opened 2026-02-03
Revise Linux installation steps in README.md

davide-ferrara:
Updated installation instructions for Linux to use curl.


gjdawson:
Why?

davide-ferrara:
It runs the installer directly with curl.
It's just convenience, is the same thing.

mcmonkey4eva:
no

PR #1277, opened 2026-02-07
Add optional PythonPath for ComfyUI self-start backends

diodiogod:
## Summary
- add an optional `PythonPath` setting to `ComfyUI Self-Starting` backends
- allow this setting to point to either a python executable or a venv root directory
- plumb the override through launch and dependency/lib detection paths
- document the new setting in the Comfy backend README

## Why
Some installs keep a valid venv outside the default `ComfyUI/venv` path. This makes self-start use that environment without requiring folder junctions/symlinks.

## Testing
- static call-site check completed
- full build was not run in this environment (`dotnet` CLI not available)

### DIFF
--- a/src/Backends/NetworkBackendUtils.cs
+++ b/src/Backends/NetworkBackendUtils.cs
@@ -225,8 +225,14 @@ public static class NetworkBackendUtils
     }
 
     /// <summary>Tries to identify the lib/site-packages/ folder that will be used for a given start script.</summary>
-    public static string GetProbableLibFolderFor(string script)
+    public static string GetProbableLibFolderFor(string script, string pythonPath = null)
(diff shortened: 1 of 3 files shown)


mcmonkey4eva:
If you want an unusual venv with custom handling, you can have your own launch script that activates the venv then launches swarm, it should just pass through properly.

For the record I don't recommend doing this, just let swarm install a stable comfy backend for you.

Next time please ask for support BEFORE spamming llm generated PRs or whatever this is.

diodiogod:
I already have too many comfyui installs, I don't want another one... OK, then won't try to do a PR anymore sorry. I'm not a dev. I'll leave your project be. Thanks.

PR #1285, opened 2026-02-12
Port history bulk image select/delete from SwarmUI_2026-02-10

Reaper176:
## Summary
- port image history bulk selection and delete controls from SwarmUI_2026-02-10
- add reusable browser entry checkbox support used by history thumbnails
- add CSS for checkbox overlays, selected entry highlighting, and bulk control layout

## Notes
- this keeps single-image delete behavior but routes it through shared delete helpers used by bulk delete

### DIFF
--- a/src/wwwroot/css/genpage.css
+++ b/src/wwwroot/css/genpage.css
@@ -251,6 +251,32 @@ body {
     display: inline-block;
     cursor: pointer;
 }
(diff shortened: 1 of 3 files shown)


mcmonkey4eva:
This is a wrong approach and needs to return to basic planning and discussion stages.

PR #1287, opened 2026-02-13
Adds support for downloading from civitai api urls

jtreminio:
Pasting any of the following URLs into the "Utilities -> Model Downloader" page will all resolve to the same path, and fetch metadata:

* https://civitai.com/models/2384321
* https://civitai.com/models/2384321?modelVersionId=2681182
* https://civitai.com/api/download/models/2681182
* https://civitai.com/api/download/models/2681182?type=Model&format=SafeTensor

This functionality also extends into the "Edit Model Metadata" modal.

### DIFF
--- a/src/wwwroot/js/genpage/utiltab.js
+++ b/src/wwwroot/js/genpage/utiltab.js
@@ -389,10 +389,8 @@ class ModelDownloaderUtil {
                 return [parts[1], null];
             }
         }
-        if (parts[0] == 'api' && parts[1] == 'download' && parts[2] == 'models' && parts.length >= 4) {
-            return [null, splitWithTail(parts[3], '?', 2)[0]];
-        }
-        if (parts[0] == 'api' && parts[1] == 'v1' && parts[2] == 'model-versions' && parts.length >= 4) {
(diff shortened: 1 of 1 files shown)


src/wwwroot/js/genpage/utiltab.js
```diff
@@ -518,10 +536,23 @@ class ModelDownloaderUtil {
...
+                    this.urlStatusArea.innerText = "URL appears to be a CivitAI model-version link, but could not resolve model metadata from Civitai API.";
+                    this.nameInput();
+                };
+                genericRequest('ForwardMetadataRequest', { 'url': `${this.civitPrefix}api/v1/model-versions/${versId}` }, (rawData) => {
+                    rawData = rawData.response;
+                    if (!rawData || !rawData.modelId) {
+                        onMetadataError();
+                        return;
+                    }
+                    loadMetadata(rawData.modelId, versId);
```
mcmonkey4eva, line 554: missed before merging, but: this feeds an int to a space that only works with a string, so this errors every time it's ran? Not sure how you tested this.
Everything seems fine with a string fix.

PR #1290, opened 2026-02-13
Reaper/save current image fix

Reaper176:


Reaper176:
sorry i was messing with my branch and messed up. soz

PR #1316, opened 2026-03-17
fix: UpdateAndRestart now explicitly fetches and checks out master br…

oxygenwalrus:
…anch

The update mechanism was failing silently because:
1. It assumed the current branch had upstream tracking
2. When on feature branches without tracking, git pull would fail without errors
3. Error handling only caught one specific error message type

Now it:
- Explicitly fetches from origin
- Checks out master branch with force flag
- Uses pull --autostash to handle local modifications
- Logs all git operations for debugging
- Catches additional error cases

Fixes: Silent failures when clicking 'update and restart' from UI


oxygenwalrus:
Ignote

PR #1317, opened 2026-03-17
Adds PostRenderTorched extension

jtreminio:
[Benchmark data available here](https://github.com/jtreminio/SwarmUI-PostRenderTorched). Up to 74% faster than original PostRender extension.

* includes a set of LUTs, and handy presets for Film Grain, Vignette, Depth Map Blur, Radial Blur
* combines all the separate features into one unified group
* bakes the (refactored) ComfyUI node into the extension itself
* no extra Python dependencies, so yay

1830001-Young American woman, smiling, looking a-6

I had a PR against the original PostRender repo but no response after over a week. Figure I might as well do it myself.

### DIFF
--- a/launchtools/extension_list.fds
+++ b/launchtools/extension_list.fds
@@ -89,10 +89,11 @@ WildcardImporter:
     - tabs
     - ui
 
-PostRenderTorched:
-    author: Juan Treminio
-    description: Adds support for a variety of useful post-processing techniques - LUTs, film grain, vignette, blur, and more. Updated version of the original PostRender extension, uses Torch, up to 74% speedup compared to the original. *Warning*: You must uninstall the old PostRender extension first before installing this one!
+PostRender:
+    author: Juan Treminio, orig. HellerCommaA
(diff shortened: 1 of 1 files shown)


mcmonkey4eva:
Considering the original is unmaintained for 2 years, do you wanna just swap the url/author label on the old one?

jtreminio:
> Considering the original is unmaintained for 2 years, do you wanna just swap the url/author label on the old one?

I didn't want to step on anyone's toes, but yeah that's fine. I'd like to keep the "Torched" portion to keep it backwards-compatible with anyone who's already got it installed.

The only question that I would need answered for this: do I swap out the node names as well, so it's a drop-in replacement for existing ComfyUI workflows? Or leave as-is and assume folks will update when they need to?

jtreminio:
Uhh, "to keep it backwards-compatible with anyone who's already got it installed." I mean if someone's already got the PostRender extension they can also install this one without any conflicts.

mcmonkey4eva:
Installing *both at the same time* seems like a bad idea, I'd think?

Issue #1269, opened 2026-02-02
Image Prompt Improvement: Drag Image onto Other Image to Replace


Deadstarrr:
### Feature Idea

When you drag an image into the prompt area it adds the image to your prompt, or adds it to the images you already have there. I would like the functionality to be added where I can drag an image onto an image that is already there to replace that image.

How it would work:
If you drag an image ONTO an image that's already above your prompt, it REPLACES that image.
If you drag an image into the empty space next to the image, it works as it does now and it adds that image into the row of images you have.

This would save a lot of time. It only takes a second or two to close the image, but if you are working with a lot of images, this really does add up.

GlenCarpenter:
@Deadstarrr I have created an extension with this functionality - https://github.com/GlenCarpenter/SwarmUIImageDropResolution

Deadstarrr:
This works REALLY well. Thank you!

PR #1341, opened 2026-04-10
add drop-to-replace functionality for image prompt drop zone

GlenCarpenter:
## add drop-to-replace functionality for image prompt drop zone

- adds drag event listeners to image prompt dropzone to watch for file replacements
- shows ~a `2px` white~ `outline: 4px solid var(--emphasis)` to indicate file will be replaced
- adds a `2px` margin to images in drop zone to prevent outline overlap

### DIFF
--- a/src/wwwroot/css/genpage.css
+++ b/src/wwwroot/css/genpage.css
@@ -1113,7 +1113,7 @@ body {
     cursor: pointer;
 }
 .alt-prompt-image-container.image-drop-replace-target .alt-prompt-image {
-    outline: 2px solid var(--box-selected-border-stronger);
+    outline: 4px solid var(--emphasis);
     border-radius: 5px;
 }
 .image-clear-button {
--- a/src/wwwroot/css/genpage.css
+++ b/src/wwwroot/css/genpage.css
@@ -1075,6 +1075,7 @@ body {
     position: relative;
     background-color: color-mix(in srgb, transparent 80%, var(--background));
     transition: background-color 0.2s ease-out;
+    margin: 0 2px;
 }
 .alt-prompt-image {
     max-height: 128px;
@@ -1110,6 +1111,10 @@ body {
     background-color: color-mix(in srgb, var(--red) 70%, black) !important;
     border: 1px solid black;
     cursor: pointer;
+}
+.alt-prompt-image-container.image-drop-replace-target .alt-prompt-image {
+    outline: 2px solid var(--box-selected-border-stronger);
+    border-radius: 5px;
 }
 .image-clear-button {
     vertical-align: middle;
--- a/src/wwwroot/js/genpage/main.js
+++ b/src/wwwroot/js/genpage/main.js
@@ -554,7 +554,9 @@ function imagePromptAddImage(file) {
     if (replaceTarget && !existingImage) {
         replaceTarget = null;
     }
-    readFileAsDataURL(file, (data) => {
+    let reader = new FileReader();
+    reader.onload = (e) => {
+        let data = e.target.result;
         if (replaceTarget && !replaceTarget.isConnected) {
             imagePromptAddImage(file);
             return;
@@ -587,7 +589,8 @@ function imagePromptAddImage(file) {
         clearButton.style.display = '';
         showRevisionInputs(true);
         genTabLayout.altPromptSizeHandle();
-    });
+    };
+    reader.readAsDataURL(file);
 }
 
 function imagePromptInputHandler() {
--- a/src/wwwroot/js/genpage/main.js
+++ b/src/wwwroot/js/genpage/main.js
@@ -537,7 +537,7 @@ function setPromptImageReplaceTarget(target) {
 }
 
 function getPromptImageDropReplaceTarget(e) {
-    if (!e || !e.dataTransfer || !e.target || !e.target.closest || uiImprover.getFileList(e.dataTransfer, e).length == 0) {
+    if (uiImprover.getFileList(e.dataTransfer, e).length == 0) {
         return null;
     }
     let target = e.target.closest('.alt-prompt-image-container');
--- a/src/wwwroot/js/genpage/main.js
+++ b/src/wwwroot/js/genpage/main.js
@@ -524,33 +524,70 @@ function autoRevealRevision() {
     }
 }
 
+let promptImageReplaceTarget = null;
+
+function setPromptImageReplaceTarget(target) {
+    if (promptImageReplaceTarget) {
+        promptImageReplaceTarget.classList.remove('image-drop-replace-target');
+    }
+    promptImageReplaceTarget = target;
+    if (promptImageReplaceTarget) {
+        promptImageReplaceTarget.classList.add('image-drop-replace-target');
+    }
+}
+
+function getPromptImageDropReplaceTarget(e) {
+    if (!e || !e.dataTransfer || !e.target || !e.target.closest || uiImprover.getFileList(e.dataTransfer, e).length == 0) {
+        return null;
+    }
+    let target = e.target.closest('.alt-prompt-image-container');
+    if (!target || !target.querySelector('.alt-prompt-image')) {
+        return null;
+    }
+    return target;
+}
+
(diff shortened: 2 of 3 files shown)


GlenCarpenter:
this is related to https://github.com/mcmonkeyprojects/SwarmUI/issues/1269

src/wwwroot/js/genpage/gentab/currentimagehandler.js
```diff
@@ -1052,7 +1052,7 @@ function setCurrentImage(src, metadata = '', batchId = '', previewGrow = false,
             canvas.toBlob(blob => {
                 let type = img.src.substring(img.src.lastIndexOf('.') + 1);
                 let file = new File([blob], imagePathClean, { type: `image/${type.length > 0 && type.length < 20 ? type : 'png'}` });
-                imagePromptAddImage(file);
+                imagePromptAddOrReplaceImage(file);
```
mcmonkey4eva: this function rename makes the commit messy, keep the original function name

src/wwwroot/js/genpage/main.js
```diff
@@ -553,6 +553,87 @@ function imagePromptAddImage(file) {
...
+    if (promptImageReplaceTarget) {
+        promptImageReplaceTarget.classList.remove('image-drop-replace-target');
+    }
+    promptImageReplaceTarget = target;
+    if (promptImageReplaceTarget) {
+        promptImageReplaceTarget.classList.add('image-drop-replace-target');
+    }
+}
+
+function isImageFileDragEvent(e) {
```
mcmonkey4eva: this function is... odd. There's existing handling for drag/drop stuff

src/wwwroot/js/genpage/main.js
```diff
@@ -553,6 +553,87 @@ function imagePromptAddImage(file) {
...
+            if (type == 'Files') {
+                return true;
+            }
+        }
+    }
+    return false;
+}
+
+function getPromptImageDropReplaceTarget(e) {
+    if (!isImageFileDragEvent(e) || !e.target || !e.target.closest) {
```
mcmonkey4eva: why would element.closest ever not exist

src/wwwroot/js/genpage/main.js
```diff
@@ -553,6 +553,87 @@ function imagePromptAddImage(file) {
...
+}
+
+function replacePromptImage(targetContainer, file) {
+    let imageObject = targetContainer ? targetContainer.querySelector('.alt-prompt-image') : null;
+    if (!imageObject) {
+        imagePromptAddImage(file);
+        return;
+    }
+    let reader = new FileReader();
+    reader.onload = (e) => {
```
mcmonkey4eva, line 558: there should not be new file reading code, there already is code for this stuff

src/wwwroot/js/genpage/main.js
GlenCarpenter: old habit I guess. It has full browser support now so this check isn't needed.

src/wwwroot/js/genpage/gentab/currentimagehandler.js
GlenCarpenter: I originally renamed the old version of `imagePromptAddImage` to `imagePromptAddImageCore` and called it in a new `imagePromptAddImage`, but this feels more descriptive to me. "adding" and "replacing" are two separate actions and this is a function that figures out which action to take. That being said it's not a hill I am going to die on so I can change it back if you feel it makes more sense.

src/wwwroot/js/genpage/main.js
GlenCarpenter: I will move this to the other event handling logic.

src/wwwroot/js/genpage/main.js
GlenCarpenter: Will refactor to use the existing file reader code.

src/wwwroot/css/genpage.css
```diff
@@ -1111,6 +1112,9 @@ body {
     border: 1px solid black;
     cursor: pointer;
 }
+.alt-prompt-image-container.image-drop-replace-target .alt-prompt-image {
+    outline: 2px solid white;
```
GlenCarpenter: @mcmonkey4eva I am going to update this to match the `:hover` state for thumbnails

```
outline: 2px solid var(--box-selected-border-stronger);
border-radius: 5px;
```

GlenCarpenter: ended up with

```
.alt-prompt-image-container.image-drop-replace-target .alt-prompt-image {
    outline: 4px solid var(--emphasis);
    border-radius: 5px;
}
```

This makes it more obvious the image is being targeted.

src/wwwroot/js/genpage/main.js
GlenCarpenter: consolidated a bunch of code

src/wwwroot/js/genpage/gentab/currentimagehandler.js
GlenCarpenter: got rid of the unnecessary functions

src/wwwroot/js/util.js
```diff
@@ -744,6 +744,18 @@ function readFileText(file, handler) {
     reader.readAsText(file);
 }
 
+/** Reads the given file as a data URL and passes the result to the given handler. Ignores null file inputs. */
+function readFileAsDataURL(file, handler) {
```
mcmonkey4eva: where did this come from now? again, there's already code that handles this part, you have no reason to be committing new file reading

mcmonkey4eva:
don't do the forcepushes btw, just do regular commits, it'll be squashed on final pull

src/wwwroot/js/genpage/main.js
```diff
@@ -524,33 +524,70 @@ function autoRevealRevision() {
...
+        promptImageReplaceTarget.classList.remove('image-drop-replace-target');
+    }
+    promptImageReplaceTarget = target;
+    if (promptImageReplaceTarget) {
+        promptImageReplaceTarget.classList.add('image-drop-replace-target');
+    }
+}
+
+function getPromptImageDropReplaceTarget(e) {
+    if (!e || !e.dataTransfer || !e.target || !e.target.closest || uiImprover.getFileList(e.dataTransfer, e).length == 0) {
```
mcmonkey4eva: again these are strange checks that don't make sense

src/wwwroot/js/util.js
GlenCarpenter: I thought this was the existing code you were referencing. What is the point of this module if we are going to create inline `FileReader`?

src/wwwroot/js/genpage/main.js
GlenCarpenter: Do you mean just the `e.target.closest` or all of the event checks?

src/wwwroot/js/util.js
mcmonkey4eva: The key rule I think you're missing here is: a PR should always seek to create the minimum possible changes needed to achieve its stated goal. Refactoring and reorganizing should not happen in a PR unless that is the focused purpose of a PR, which this is not.
There's already code that handles reading a file in the correct place, so you should not be modifying it at all here, just redirecting where the resultant image goes.

src/wwwroot/js/genpage/main.js
mcmonkey4eva: The closest one is the obvious problem, but I imagine most of the checks here can never fail, and so don't need to be checked

src/wwwroot/js/util.js
GlenCarpenter: Makes sense, I see what you're saying. I'll restore the old FileReader and keep all the new code localized to the existing `imagePromptAddImage`

src/wwwroot/js/genpage/main.js
GlenCarpenter: this should be fixed now

src/wwwroot/js/util.js
GlenCarpenter: FileReader has been restored

PR #1344, opened 2026-04-14
Api backends extension

kalebbroo:
Add the API Backends Extension to the Extensions tab


mcmonkey4eva:
get rid of ci-test, and add 'paid'

launchtools/extension_list.fds
```diff
@@ -277,3 +287,4 @@ AceStepFun:
     ci-test: true
     tags:
     - tools
+
```
mcmonkey4eva: wat

PR #1347, opened 2026-04-14
Adds Ernie support

jtreminio:
1222003-A young, African woman holds a large, wh-3

CleanShot 2026-04-14 at 12 25 00

Confirmed no model metadata mismatches against suite of models.


jtreminio:
I'm writing the docs now!

kwal559:
Going to support the prompt enhancer ? or just go with the comfy workflow?  It's odd that the text encoder and the prompt enhancer are 2 individual mistral3 models..

mcmonkey4eva:
Dedicated LLM prompt upsampling (separate from any given image model) is a planned feature, and currently available via the _MagicPrompt_ extension

PR #1352, opened 2026-04-18
Add Docker AMD Support

bigcletus:
I made a new swarm docker-compose.yml file, new StandardDockerfile.docker, and new ReadMe with instructions. I have only tested the distribution on a cluster of 1 computer, will test on more another time. This should improve scalability while adding AMD GPU support for those who want to run SwarmUI in Docker.


mcmonkey4eva:
I think this could just be a couple comments/args on the existing scripts, and doesn't need to be whole new ones.

Also get rid of the dockerswarm stuff 0.o

bigcletus:
Will make additional changes when I wake up, bedtime for me

bigcletus:
How do you feel about my recent changes? To put it briefly:

Added an extra commandline arg to launch-standard-docker.sh (You can now put "amd" or "nvidia" so long as fixch comes last)
Removed the Docker Swarm Support :( I'm personally still going to use it in my own environment
Kept the custom AMDGPU-StandardDockerfile.docker unchanged, since it took forever to get working and I don't have the courage to fool with it since it works. I could theoretically thin out the deps since I got it working and a few are redundant now.

bigcletus:
Oh, and the swarm images generated are called either swarmui-rocm or swarmui-nvidia. They have been implemented into the launch script properly

mcmonkey4eva:
Stop doing _more_ things in one PR. Do the minimum necessary for the PR's intended purpose, which is AMD support for docker.

Anything else can be a separate followup discussion.

bigcletus:
Sorry, I'm new to this whole github thing. Can you tell I'm an amateur? Lol

mcmonkey4eva:
Oh, you hand-merged the last fix commit instead of using a standard git merge, this prevents the PR from being merged. (If you're unfamiliar with git, you can also close and rebuild the PR if that's easier than fixing a git merge)

PR #1353, opened 2026-04-19
Add AMD GPU Docker Support

bigcletus:
Updated version of SwarmUI with AMD Support for the launch-standard-docker.sh script.


mcmonkey4eva:
As written previously:
- Please remove the entire separate file
- Please remove the container name changes
- In general remove all non-essential changes

bigcletus:
The dockerfile you use pulls from microsoft's .net registry, which does not support ROCm. I'll work on combining the two with minimal changes.

mcmonkey4eva:
that doesn't support nvidia either, the dockerfile itself isn't meant to be a GPU install

PR #1354, opened 2026-04-19
Add AMD GPU Support to Dockerfile

bigcletus:
I have tested this on a 9070XT and it works fine. I do not have an nvidia GPU to test on though, so I can't promise nothing broke.


mcmonkey4eva:
this just runs some amd install in the docker blindly for all users

bigcletus:
Alright, so for whatever reason ROCm works out-of-the-box for me now. I guess I had something misconfigured on my host and this whole process fixed it. Sorry for wasting your time.

PR #1345, opened 2026-04-14
Adding Image Comparison

jtreminio:
* **I needed AI assistance for the CSS portion**. There was no way I was going to figure that out for myself.
* I needed AI assistance for the maths in the JS. The rest of it is mine, including all bugs.
* I've tried to reuse as much code as possible from `ImageFullViewHelper` for the new `ImageCompareHelper`.

https://github.com/user-attachments/assets/872dc753-da53-4c28-a22a-0c9235b0464f

PR #1360, opened 2026-04-30
Adds multi-select support

jtreminio:
* works across batch and history containers, synced to each other.
* can be easily hooked into (see #1345)
* when enabled, clicking a card description selects the card

https://github.com/user-attachments/assets/02d4d014-4a83-472c-910f-d9b3b3c18479


src/wwwroot/js/genpage/gentab/outputhistory.js
```diff
@@ -247,6 +254,62 @@ function selectOutputInHistory(image, div) {
 
 let imageHistoryBrowser = new GenPageBrowserClass('image_history', listOutputHistoryFolderAndFiles, 'imagehistorybrowser', 'Thumbnails', describeOutputFile, selectOutputInHistory,
     `<label for="image_history_sort_by">Sort:</label> <select id="image_history_sort_by"><option>Name</option><option>Date</option></select> <input type="checkbox" id="image_history_sort_reverse"> <label for="image_history_sort_reverse">Reverse</label> &emsp; <input type="checkbox" id="image_history_allow_anims" checked autocomplete="off"> <label for="image_history_allow_anims">Allow Animation</label>`);
+imageHistoryBrowser.enableBrowserMultiSelect = true;
+imageHistoryBrowser.keepBrowserMultiSelectKeyAfterPrune = function(key, namesInCurrentList) {
```
mcmonkey4eva: match existing patterns for function declarations

src/wwwroot/js/genpage/gentab/currentimagehandler.js
```diff
@@ -436,6 +436,11 @@ function toggleSeparateBatches() {
 
 function clickImageInBatch(div) {
     let imgElem = div.getElementsByTagName('img')[0];
+    let multiSelectKey = getImageFullSrc(div.dataset.src);
```
mcmonkey4eva: there shouldn't be any multiselect handling outside of the browser class like this

src/wwwroot/js/genpage/gentab/models.js
```diff
@@ -717,7 +718,7 @@ class ModelBrowserWrapper {
             }, can_multi: true }];
         }
         let isStarred = this.isStarred(model.data.name);
-        let starButton = { label: isStarred ? 'Unstar' : 'Star', onclick: () => { this.toggleStar(model.data.name); } };
+        let starButton = { label: isStarred ? 'Unstar' : 'Star', onclick: () => { this.toggleStar(model.data.name); }, can_multi: true };
```
mcmonkey4eva: no

src/wwwroot/js/genpage/gentab/outputhistory.js
```diff
@@ -72,7 +72,8 @@ function buttonsForImage(fullsrc, src, metadata, isCurrentImage = false) {
             className: (metadata && metaParsed.is_starred) ? ' star-button button-starred-image' : ' star-button',
             onclick: (e) => {
                 toggleStar(fullsrc, src);
-            }
+            },
+            can_multi: true
```
mcmonkey4eva: no

src/wwwroot/js/genpage/gentab/outputhistory.js
```diff
@@ -167,6 +168,11 @@ function buttonsForImage(fullsrc, src, metadata, isCurrentImage = false) {
                     if (div) {
                         removeImageBlockFromBatch(div);
                     }
+                    if (imageHistoryBrowser.enableBrowserMultiSelect) {
```
mcmonkey4eva: this is questionably placed aaand also seems to be a mixup between the 'enable' and 'active' bools?

src/wwwroot/js/genpage/gentab/outputhistory.js
```diff
@@ -247,6 +254,62 @@ function selectOutputInHistory(image, div) {
 
 let imageHistoryBrowser = new GenPageBrowserClass('image_history', listOutputHistoryFolderAndFiles, 'imagehistorybrowser', 'Thumbnails', describeOutputFile, selectOutputInHistory,
     `<label for="image_history_sort_by">Sort:</label> <select id="image_history_sort_by"><option>Name</option><option>Date</option></select> <input type="checkbox" id="image_history_sort_reverse"> <label for="image_history_sort_reverse">Reverse</label> &emsp; <input type="checkbox" id="image_history_allow_anims" checked autocomplete="off"> <label for="image_history_allow_anims">Allow Animation</label>`);
+imageHistoryBrowser.enableBrowserMultiSelect = true;
+imageHistoryBrowser.keepBrowserMultiSelectKeyAfterPrune = function(key, namesInCurrentList) {
+    if (namesInCurrentList.has(key)) {
+        return true;
+    }
+    let currentImageBatchDiv = getRequiredElementById('current_image_batch');
+    for (let candidate of currentImageBatchDiv.querySelectorAll('.image-block:not(.image-block-placeholder)')) {
```
mcmonkey4eva: this is an odd query

src/wwwroot/js/genpage/gentab/presets.js
```diff
@@ -626,18 +626,18 @@ function listPresetFolderAndFiles(path, isRefresh, callback, depth) {
 function describePreset(preset) {
     let buttons = [
         { label: 'Toggle', onclick: () => selectPreset(preset) },
-        { label: 'Direct Apply', onclick: () => applyOnePreset(preset.data) },
-        { label: preset.data.is_starred ? 'Unstar' : 'Star', onclick: () => togglePresetStar(preset) },
+        { label: 'Direct Apply', onclick: () => applyOnePreset(preset.data), can_multi: true },
+        { label: preset.data.is_starred ? 'Unstar' : 'Star', onclick: () => togglePresetStar(preset), can_multi: true },
```
mcmonkey4eva: no

src/wwwroot/js/genpage/gentab/presets.js
```diff
@@ -626,18 +626,18 @@ function listPresetFolderAndFiles(path, isRefresh, callback, depth) {
...
         { label: 'Edit Preset', onclick: () => editPreset(preset.data) },
         { label: 'Duplicate Preset', onclick: () => duplicatePreset(preset.data) },
-        { label: 'Export Preset', onclick: () => exportOnePresetButton(preset.data) },
+        { label: 'Export Preset', onclick: () => exportOnePresetButton(preset.data), can_multi: true },
```
mcmonkey4eva: this can't function natively and needs separate impl

src/wwwroot/js/genpage/helpers/browsers.js
```diff
@@ -107,6 +107,11 @@ class GenPageBrowserClass {
         this.runAfterUpdate = [];
         this.refreshHandler = (callback) => callback();
         this.checkIsSmall();
+        this.enableBrowserMultiSelect = false;
```
mcmonkey4eva: should probably be `allowMultiSelect`

src/wwwroot/js/genpage/helpers/browsers.js
```diff
@@ -107,6 +107,11 @@ class GenPageBrowserClass {
...
+        this.multiSelectActive = false;
+        this.multiSelectedKeys = new Set();
```
mcmonkey4eva: this is liable to lead to state tracking inconsistencies, as jank as it feels putting class or .dataset's on the block divs is probably the stablest way to track

src/wwwroot/js/genpage/helpers/browsers.js
```diff
@@ -467,6 +479,20 @@ class GenPageBrowserClass {
                 textBlock.tabIndex = 0;
                 textBlock.innerHTML = desc.description;
                 div.appendChild(textBlock);
+                div.addEventListener('click', (e) => {
```
mcmonkey4eva: don't need a separate click handler, the select handler should cover it

src/wwwroot/js/genpage/gentab/models.js
jtreminio: Are you objecting to the `, can_multi: true` flag itself (how I've defined it, for example), or objecting to Star/Unstar action being available for multi-select action?

mcmonkey4eva: you editing the way star is handled for multi, which is already covered properly for images, and presets/models should handle it the same as images

mcmonkey4eva: Probably the best move is limit the PR to browser itself & images, preset/models can be separate

aka the universal rule of PR contributing: minimize the scope of edits per individual PR

src/wwwroot/js/genpage/gentab/outputhistory.js
```diff
@@ -65,7 +65,15 @@ function buttonsForImage(fullsrc, src, metadata, isCurrentImage = false) {
...
-        let metaParsed = JSON.parse(metadata);
+        let getMeta = (metadata) => metadata ? (JSON.parse(metadata) || {}) : {};
+        let metaParsed = getMeta(metadata);
+        let isStarred = (e) => {
+            let currentMeta = e && e.dataset ? getMeta(e.dataset.metadata) : {};
+            if (Object.keys(currentMeta).length == 0) {
+                currentMeta = metaParsed;
+            }
+            return currentMeta.is_starred;
+        };
```
jtreminio, line 76: There was a bug in your original implementation:

1) metadata would initially not exist on an item, causing exception on `let metaParsed = JSON.parse(metadata);`
2) metadata would be stale on reads below. Clicking "Enable Starred" would add the needed metadata and also star the items, but then clicking "Disable(d) Starred" would still see stale metadata and not unstar items
3) Fixed speling

src/wwwroot/js/genpage/helpers/browsers.js
```diff
@@ -466,6 +474,12 @@ class GenPageBrowserClass {
                 let textBlock = createDiv(null, 'model-descblock');
                 textBlock.tabIndex = 0;
                 textBlock.innerHTML = desc.description;
+                textBlock.addEventListener('click', (e) => {
```
mcmonkey4eva: What's with all these new click listeners?

src/wwwroot/js/genpage/helpers/browsers.js
```diff
@@ -783,4 +842,222 @@ class GenPageBrowserClass {
...
+    /**
+     * Returns multi-select items.
+     */
+    getMultiSelectedItems() {
+        if (!this.contentDiv) {
+            return [];
+        }
+        let items = [];
+        for (let child of this.contentDiv.children) {
+            if (child.dataset && child.dataset.name && child.classList.contains('browser-multiselect-item-selected')) {
```
mcmonkey4eva: querySelectorAll should do this quick n simple

src/wwwroot/js/genpage/helpers/browsers.js
jtreminio: This one is necessary. I couldn't ~need~ get multi-select click to select a card if I clicked the description otherwise.

mcmonkey4eva: correct

src/wwwroot/js/genpage/helpers/browsers.js
```diff
@@ -783,4 +836,206 @@ class GenPageBrowserClass {
...
+        for (let file of this.lastFiles) {
+            if (selectedNames.has(file.name)) {
+                out.push(file);
+            }
+        }
+        return out;
+    }
+
+    /**
+     * Labels for bulk actions shared by every selected item, respecting `can_multi` / `multi_only`.
```
mcmonkey4eva: This feels like AI slop. Misinterpreted keywords and then proudly documented respect for the misunderstood keyword.

src/wwwroot/js/genpage/helpers/browsers.js
```diff
@@ -783,4 +836,206 @@ class GenPageBrowserClass {
...
+     */
+    getCommonMultiSelectActionLabels() {
+        let files = this.getMultiSelectedFiles();
+        if (files.length == 0) {
+            return [];
+        }
+        let eligiblePerFile = [];
```
mcmonkey4eva, line 928: actually wtf is this whole function? Why was this built like this at all??? wat.

mcmonkey4eva:
Issues from testing:
- Multi-select overlaps with regular select. The image I have selected *looks like* it is selected for multi-select but isn't actually. This is weird. Probably kill any singleselect classes.
- After testing, I see some value to having a set like you did before: state is currently killed on any refresh, so need to persist past that. I'd minimize the set handling to be exclusively *just* a tool to reapply the selected class after a refresh (grab a set of what elems have the class before, apply back after, call it a day)

src/wwwroot/js/genpage/helpers/browsers.js
jtreminio: To be fair, this PR has gone through several major revisions after feedback. I try to keep comments up to date.

It does point to my initial PR diffs probably being overly large if I sometimes mislabel a function, which is something I'm actively working on for the SwarmUI project.

src/wwwroot/js/genpage/helpers/browsers.js
jtreminio: I'm reducing the diff here, but this function actually does something important - it gates specific multi-select actions on filetypes the user selects.

A user can select any number of image, video, and audio files in the history browser and select an action like Star, or Delete.

However, it makes less sense for a user to select an image + audio file and allow them to do a [media comparison](https://github.com/mcmonkeyprojects/SwarmUI/pull/1345).

The current implementation just iterates through each selected file and removes actions that cannot be applied to the entire set. Slightly wordy, I can reduce the diff without getting too clever, but I would push back strongly on this being removed completely.

jtreminio: "image + audio" comparison is stupid and I would instead say an image + video comparison wouldn't make much sense. However, it would make more sense than image+audio.

I can _maybe_ see a scenario where a user wants to popup an image and a video into a comparison modal to eyeball a bit? Not much of a scenario but if you want this code removed now I'm open to it.

src/wwwroot/js/genpage/helpers/browsers.js
```diff
@@ -755,6 +807,11 @@ class GenPageBrowserClass {
             });
         }
         else {
+            if (this.preservedMultiSelect.size == 0) {
```
mcmonkey4eva: This set is used and cleared directly within one function, so presumably can safely be function-local rather than on the instance ctor?

jtreminio: _kind of_. Doing a filter or depth change doesn't need an ~object~ class instance variable, but the browser refresh button nukes everything. I couldn't get state to stick otherwise. I'm pushing one small change cleaning up tiny odds and ends but unless you have a suggested alternative method, this is the way I could get it to work.

Note that if a user clicks the Refresh button twice very quickly, state is still lost. I was going to add a timeout but then PR starts touching things unrelated to core feature.

src/wwwroot/js/genpage/helpers/browsers.js
```diff
@@ -783,4 +848,180 @@ class GenPageBrowserClass {
...
+        let files = this.getMultiSelectedFiles();
+        if (files.length == 0) {
+            return [];
+        }
+        let eligiblePerFile = [];
+        for (let file of files) {
+            let desc = this.describe(file);
+            let labels = new Set();
+            for (let button of desc.buttons) {
+                if (button.can_multi && (button.max_selected == null || files.length <= button.max_selected)) {
```
jtreminio, line 933: TBH I'm not a fan of the change you made here. The previous split `if` blocks were wordy and we could have merged a few of them, but this `if` now feels overloaded.

PR #1361, opened 2026-05-01
Allows using a video for ControlNet preprocessor, keeps first frame

jtreminio:
CleanShot 2026-05-01 at 18 12 11


mcmonkey4eva:
This is wrong impl, this is just not using the video.

PR #1367, opened 2026-05-03
Expose the context parameter during creation of TensorRT engines

jakstein:
A simple tweak that exposes previously hardcoded context parameter (was hardcoded to 1) when creating TensorRT engines. Now users will get to actually select how big context support they want in the dialog, along with other params.
This allows the created engines to actually work with prompts exceeding 75 tokens.
If user inputs 75 in dialog, the value is set to 1, if 150 then 2, so on.


mcmonkey4eva:
Entirely redundant imo considering the unmaintained and replaced status of tensorrt, but sure why not

PR #1369, opened 2026-05-05
Add support for extension private deps

jtreminio:
The way this works for me is, in my extension's `.csproj` file:

```xml
  <ItemGroup>
    <Reference Include="ComfyTyped">
      <HintPath>lib/ComfyTyped.dll</HintPath>
    </Reference>
  </ItemGroup>
```

I've added the ComfyTyped dep to two of my extensions, one with it fully baked in and used, the other with a very small footprint, swarmui builds fine.

This PR's purpose is for each extension to define its own private dependency, where ExtensionA and ExtensionB both depend on LibA but might be different versions.

The simpler solution that does not guard against multiple extensions using the same dependency that might have differing versions is changing `return Assembly.LoadFile` to `return Assembly.LoadFrom` here:

* https://github.com/mcmonkeyprojects/SwarmUI/blob/master/src/Core/ExtensionsManager.cs#L184
* https://github.com/mcmonkeyprojects/SwarmUI/blob/master/src/Core/ExtensionsManager.cs#L198

Full disclosure as mentioned in Discord but important to note: This PR was created using Claude because C# build and dependency management confuses the heck out of me. I know the two proposed solutions work because I have a full test suite for my extensions, SwarmUI builds, and my extension with its private dependency works great.

### DIFF
--- a/src/Core/ExtensionsManager.cs
+++ b/src/Core/ExtensionsManager.cs
@@ -12,8 +12,7 @@ public class SwarmExtensionLoadContext : AssemblyLoadContext
 {
     private readonly string ExtensionDir;
 
-    public SwarmExtensionLoadContext(string name, string extensionDir)
-        : base(name, isCollectible: false)
+    public SwarmExtensionLoadContext(string name, string extensionDir) : base(name, isCollectible: false)
     {
         ExtensionDir = extensionDir;
     }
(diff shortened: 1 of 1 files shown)


src/Core/ExtensionsManager.cs
```diff
@@ -45,6 +72,15 @@ public static HtmlString HtmlTags(string[] tags)
     /// <summary>List of known online available extensions.</summary>
     public List<ExtensionInfo> KnownExtensions = [];
 
+    /// <summary>Per-extension load contexts, keyed by extension DLL name. Each extension gets its own <see cref="SwarmExtensionLoadContext"/> so private dependencies stay isolated from other extensions.</summary>
+    public ConcurrentDictionary<string, SwarmExtensionLoadContext> ExtensionLoadContexts = new();
```
mcmonkey4eva: Why is this a retained dict on its own? It doesn't look like there's any reuse, just a single grab on load

jtreminio: Fair critique. I've reduced this further.

PR #1387, opened 2026-05-20
Adds `SwarmVideoResampleFPS`; resamples controlnet preview videos

jtreminio:
Adds new Comfy node `SwarmVideoResampleFPS`, which allows resampling videos FPS to a target FPS.

CleanShot 2026-05-19 at 22 28 42@2x


src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmVideo.py
```diff
@@ -0,0 +1,137 @@
...
+import math
+import torch
+from comfy_api.latest import io
+
+logger = logging.getLogger(__name__)
+
+
+class SwarmVideoResampleFPS(io.ComfyNode):
+    MIN_FPS: float = 1.0
+    MAX_FPS: float = 120.0
```
mcmonkey4eva: general rule for comfy inputs range limits is go way wider than seems sane. Never know when somebody's gonna complain about the lat of support for their 640 fps input video.

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmVideo.py
```diff
@@ -0,0 +1,137 @@
...
+    @classmethod
+    def define_schema(cls) -> io.Schema:
+        return io.Schema(
+            node_id="SwarmVideoResampleFPS",
+            display_name="Swarm Video Resample FPS",
+            category="SwarmUI/video",
+            description="Resample a video from fps_in to fps_out while preserving total duration.",
+            inputs=[
+                io.Image.Input(
+                    "images",
```
mcmonkey4eva: very unnecessary linesplits

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmVideo.py
```diff
@@ -0,0 +1,137 @@
...
+            resampled = cls._sample_linear(images, source_positions)
+
+        logger.info(
+            "SwarmVideoResampleFPS: %d frames @ %s fps -> %d frames @ %s fps (%s)",
+            frame_count_in, fps_in, frame_count_out, fps_out, method,
+        )
+        return io.NodeOutput(resampled, float(fps_out))
+
+    @classmethod
+    def _source_positions(cls, frame_count_out: int, fps_in: float, fps_out: float, device: torch.device) -> torch.Tensor:
```
mcmonkey4eva: there's more function body here than code, for a function used only once. This is "looks proper" code that's actually quite messy.

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmVideo.py
```diff
@@ -0,0 +1,137 @@
...
+        frame_count_out = max(1, round(duration_sec * fps_out))
+        source_positions = cls._source_positions(frame_count_out, fps_in, fps_out, images.device)
+
+        if method == cls.METHOD_NEAREST:
+            resampled = cls._sample_nearest(images, source_positions)
+        else:
+            resampled = cls._sample_linear(images, source_positions)
+
+        logger.info(
+            "SwarmVideoResampleFPS: %d frames @ %s fps -> %d frames @ %s fps (%s)",
```
mcmonkey4eva: why would you do this in a world where f strings exist

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmVideo.py
jtreminio: Copied from the [CreateVideo node](https://github.com/Comfy-Org/ComfyUI/blob/master/comfy_extras/nodes_video.py#L130). I'll bump it to `1000`? Doesn't seem like it would be valid down the chain though.

edit: ah, for input. For output the CreateVideo node would max out at 120fps.

Do you want a `1000` max input, and 120 max output?

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmVideo.py
jtreminio: Force of habit from enterprise logging libs 👍

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmVideo.py
jtreminio: Changed, but imho split lines for more than 2 parameters is vastly more readable than having single lines except for something that should go in a single line:

```python
io.Image.Input("images", tooltip="The images to resample."),
io.Float.Input("fps_in", min=cls.MIN_FPS, max=cls.MAX_FPS, step=cls.STEP_FPS, tooltip="Source frame rate."),
io.Float.Input("fps_out", default=cls.DEFAULT_FPS_OUT, min=cls.MIN_FPS, max=cls.MAX_FPS, step=cls.STEP_FPS, tooltip="Target frame rate."),
io.Combo.Input("method", options=[cls.METHOD_LINEAR, cls.METHOD_NEAREST], default=cls.METHOD_LINEAR,
    tooltip=(
        "linear: each output frame is a linear blend of the two source frames bracketing its timestamp. "
        "Equivalent to ffmpeg's framerate filter. Slightly more expensive; avoids the duplicated-frame artifact. "
        "See https://ffmpeg.org/ffmpeg-filters.html#framerate\n"
        "nearest: each output frame is the source frame closest in time. "
        "Equivalent to ffmpeg's fps filter. Cheap; can produce visible judder on pans. "
        "See https://ffmpeg.org/ffmpeg-filters.html#fps-1"
    ),
),
```

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmVideo.py
jtreminio: I'm going to ask you to reconsider this specific comment, but changing it pending.

`_source_positions()` could be boiled down to a single line:

```python
return torch.arange(frame_count_out, dtype=torch.float64, device=device) / fps_out * fps_in
```

but then anyone not familiar with this functionality will read this as a foreign language. Yes it makes sense to you and I, but I'm going to take a minute or two in a year to understand what's going on. Anyone new to this whole world of 1girl will not understand what this line is actually doing.

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmVideo.py
```diff
@@ -0,0 +1,102 @@
...
+import torch
+from comfy_api.latest import io
+
+logger = logging.getLogger(__name__)
+
+
+class SwarmVideoResampleFPS(io.ComfyNode):
+    MIN_FPS: float = 1.0
+    MAX_FPS_IN: float = 1000.0
+    MAX_FPS_OUT: float = 120.0
```
mcmonkey4eva: this split is silly, just allow [0.01, 99999] there's no reason to have strict range limits anywhere here except creating future edge case bugs to fix

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmVideo.py
```diff
@@ -0,0 +1,102 @@
...
+                io.Image.Input("images", tooltip="The images to resample."),
+                io.Float.Input("fps_in", min=cls.MIN_FPS, max=cls.MAX_FPS_IN, step=cls.STEP_FPS, tooltip="Source frame rate."),
+                io.Float.Input("fps_out", default=cls.DEFAULT_FPS_OUT, min=cls.MIN_FPS, max=cls.MAX_FPS_OUT, step=cls.STEP_FPS, tooltip="Target frame rate."),
+                io.Combo.Input("method", options=[cls.METHOD_LINEAR, cls.METHOD_NEAREST], default=cls.METHOD_LINEAR,
+                    tooltip=(
+                        "linear: each output frame is a linear blend of the two source frames bracketing its timestamp. "
+                        "Equivalent to ffmpeg's framerate filter. Slightly more expensive; avoids the duplicated-frame artifact. "
+                        "See https://ffmpeg.org/ffmpeg-filters.html#framerate\n"
+                        "nearest: each output frame is the source frame closest in time. "
+                        "Equivalent to ffmpeg's fps filter. Cheap; can produce visible judder on pans. "
```
mcmonkey4eva: weird line splits here

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmVideo.py
```diff
@@ -0,0 +1,102 @@
...
+                        "See https://ffmpeg.org/ffmpeg-filters.html#fps-1"
+                    ),
+                ),
+            ],
+            outputs=[
+                io.Image.Output("images"),
+                io.Float.Output("fps"),
```
mcmonkey4eva: unnecessary line splits

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmVideo.py
mcmonkey4eva:
```
writing(
               like,
            this,
            is
            awful
)
```
and I can't understand why you'd do it. It's significantly less readable to my eyes.
`io.Image.Input("images", tooltip="The images to resample."),` is "Here's an input, it's images type with this tooltip text", perfect and simple
whereas the multiline split goes into the mode of "here's a wall of unstructured data, skip it unless/until you're willing to spend 5 minutes parsing a wall of data"

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmVideo.py
mcmonkey4eva: I want this node to not be the node in the chain that breaks things and need to be edited to accommodate.

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmVideo.py
mcmonkey4eva: Nobody who is lacking the tools to decipher what `source_positions` means or what a `torch.arange` is going to understand either version of this code, before or after.

You can leave a `# comment` explaining in one line what it is, maybe an example of inputs -> outputs, that much can help clarify.

src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
```diff
@@ -471,6 +471,7 @@ public WGNodeData LoadImage(ImageFile img, string param, bool resize, string nod
                     {
                         ["video"] = NodePath(result, 0)
                     });
+                    NodeHelpers["video_components_split"] = splitNode;
```
mcmonkey4eva: This method of forwarding data is liable to get mixed up from a different load call in the same workflowgen

jtreminio: Should I add a new property to `WGNodeData`? `SourceFPSNode` or similar?

mcmonkey4eva: Probably change `public int? FPS = null;` to a `JToken` that can either be a raw value or a node path ref

src/BuiltinExtensions/ComfyUIBackend/ExtraNodes/SwarmComfyCommon/SwarmVideo.py
```diff
@@ -0,0 +1,94 @@
...
+                io.Float.Input("fps_in", min=cls.MIN_FPS, max=cls.MAX_FPS, step=cls.STEP_FPS, tooltip="Source frame rate."),
+                io.Float.Input("fps_out", default=cls.DEFAULT_FPS_OUT, min=cls.MIN_FPS, max=cls.MAX_FPS, step=cls.STEP_FPS, tooltip="Target frame rate."),
+                io.Combo.Input("method", options=[cls.METHOD_LINEAR, cls.METHOD_NEAREST], default=cls.METHOD_LINEAR,
+                    tooltip=(
+                        "linear: each output frame is a linear blend of the two source frames bracketing its timestamp. Equivalent to ffmpeg's framerate filter. Slightly more expensive; avoids the duplicated-frame artifact. See https://ffmpeg.org/ffmpeg-filters.html#framerate\n"
+                        "nearest: each output frame is the source frame closest in time. Equivalent to ffmpeg's fps filter. Cheap; can produce visible judder on pans. See https://ffmpeg.org/ffmpeg-filters.html#fps-1"
+                    ),
+                ),
+            ],
+            outputs=[io.Image.Output("images"), io.Float.Output("fps")],
```
mcmonkey4eva: second out is redundant, just returns an input

src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
```diff
@@ -1090,6 +1090,18 @@ bool getBestFor(string phrase)
...
+                        {
+                            string resampleNode = g.CreateNode("SwarmVideoResampleFPS", new JObject()
+                            {
+                                ["images"] = imageNodeActual.Path,
+                                ["fps_in"] = imageNodeActual.FPS,
+                                ["fps_out"] = 24.0,
+                                ["method"] = "linear"
+                            });
+                            imageNodeActual = imageNodeActual.WithPath([resampleNode, 0]);
+                            imageNodeActual.FPS = 24;
```
mcmonkey4eva: hardcoded 24 is weird

src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
```diff
@@ -1090,6 +1090,18 @@ bool getBestFor(string phrase)
                             ["scale_method"] = "lanczos"
                         });
                         imageNodeActual = imageNodeActual.WithPath([multipleOf8, 0]);
+                        if (imageNodeActual.DataType == WGNodeData.DT_VIDEO && imageNodeActual.FPS is not null)
```
mcmonkey4eva: wrong order of operations, this should be *before* preproc

src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorSteps.cs
```diff
@@ -2124,7 +2136,7 @@ WGNodeData appendAudio(WGNodeData combinedAudio, WGNodeData nextAudio)
                     JArray interp = g.DoInterpolation(g.CurrentMedia.Path, method, mult);
                     g.CurrentMedia = g.CurrentMedia.WithPath(interp);
                     videoFps *= mult;
-                    g.CurrentMedia.FPS = videoFps;
+                    g.CurrentMedia.FPS = videoFps.HasValue ? videoFps.Value : null;
```
mcmonkey4eva: just bad edit

PR #1390, opened 2026-05-23
Adds Lens support

jtreminio:
1251001-A high-action manga-style oil painting i-2

The `CFGNorm` node helps prevent super-plastic skin.

CleanShot 2026-05-23 at 12 35 53@2x

Note: I left the changes in `docs/Model Support.md` as minimal as possible so you can add your own flair. Tested censoring, it's uncensored but like z-image has little knowledge about genitals.

Default dimensions 1440 per workflow and [inference script](https://github.com/microsoft/Lens/blob/main/inference.py#L38). I noticed text has more misspellings at 1440 than 1024.

Depends on https://github.com/Comfy-Org/ComfyUI/pull/14077

### DIFF
--- a/docs/Model Support.md
+++ b/docs/Model Support.md
@@ -20,6 +20,7 @@
 [Anima](#anima) | DiT | 2026 | Circlestone Labs | 2B | WTF | Modern, very small, decent for anime |
 [ERNIE](#ernie) | DiT | 2026 | Baidu | 8B | Minimal | Modern, intelligent, good quality, fast |
 [HiDream O1](#hidream-o1) | "Pixel UiT" | 2026 | HiDream | 8B | Minimal | Modern, intelligent, fast, decent quality |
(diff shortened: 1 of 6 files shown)


jtreminio:
One fun thing to note, Lens (non-turbo, non-base) wants Euler Ancestral.

Here's Lens @ euler/normal, 20 steps, 5 cfg:

1329001-High definition photo of an orange cat w
1329001-High definition photo of an orange cat w-1
1329001-High definition photo of an orange cat w-2
1329001-High definition photo of an orange cat w-3

Here's the same with euler_a:

1330001-High definition photo of an orange cat w
1330001-High definition photo of an orange cat w-1
1330001-High definition photo of an orange cat w-2
1330001-High definition photo of an orange cat w-3

Aside from it refusing to spell "stupid" (even after ~50 attempts), small letters are messed up in euler, but look fine in euler_a.

PR #1407, opened 2026-06-05
Shows Image parameters (side view only) in columnar form for easier readability

richardtallent:
When parameter data is viewed *beside* an image, the keys and values are easily confused.

This CSS tweak improves readability showing them in pseudo-columns. It also moves the `:` to the key's span.

Reference / screenshot:

https://discord.com/channels/1243166023859961988/1243185862234210389/1510800649011593367

Bottom mode display is not affected -- an enhancement could be made there, but would likely require adjusting the markup to use CSS grids.


mcmonkey4eva:
i don't really love it but eh why not

PR #1413, opened 2026-06-10
Use Qwen3VL for ideogram4 instead of Qwen3

BetaDoggo:
Right now Swarm downloads and defaults to the regular qwen3 model, while the [official docs](https://github.com/ideogram-oss/ideogram4/blob/main/docs/model_architecture.md) and [comfy-org distribution](https://huggingface.co/Comfy-Org/Ideogram-4/tree/main/text_encoders) both use the vl variant.

I chose the fp8 scaled version instead of the nvfp4 version because I have serious concerns about TE quality degradation in this model. I know nvidia will say otherwise because they want to sell their new gpus with nvfp4 acceleration.

In my own testing I've found that the vl variant produces more coherent outputs.
# with qwen3:
image

# with qwen3vl:
image

## Much Better!

### DIFF
--- a/src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorModelSupport.cs
+++ b/src/BuiltinExtensions/ComfyUIBackend/WorkflowGeneratorModelSupport.cs
@@ -596,6 +596,11 @@ public partial class WorkflowGenerator
             return RequireClipModel("qwen_3_8b.safetensors", "https://huggingface.co/Comfy-Org/flux2-klein-9B/resolve/main/split_files/text_encoders/qwen_3_8b_fp4mixed.safetensors", "bbf16f981d98e16d080c566134814c4e9f6aadd0d0e1383c60bc44ba939d760d", T2IParamTypes.QwenModel);
         }
 
+        public string GetQwen3Vl8bModel()
+        {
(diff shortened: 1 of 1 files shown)


mcmonkey4eva:
`I chose the fp8 scaled version instead of the nvfp4 version because I have serious concerns about TE quality degradation in this model. I know nvidia will say otherwise because they want to sell their new gpus with nvfp4 acceleration.`
Concerns, or evidence?

All testing thus far with existing mdoels has shown that nvfp4 is very near to pixel identical to fp8 for the textencoder, and it's *possible* ideogram is different but I wouldn't overly expect it

BetaDoggo:
I'm not certain, and I don't have anywhere near the compute to verify, but I feel like the effect must be greater than with other models since prompts can use 20-100x the number of tokens.

I can't produce enough samples to properly evaluate degradation on my hardware but the compositional difference seems larger than what I've seen with previous models:
|fp8|nvfp4|
|---|---|
|Untitled2|BrKj4rFa|

(same seed of course)
This is probably the worst case scenario but I needed a quick (3 minute each) example


I guess nvfp4 is probably good enough to be worth saving 4GB in the default workflow, though I still oppose it on a spiritual level.

mcmonkey4eva:
Yeah okay, I did my own testing and see the same, so I put the fp8 in. I guess the more complex json prompt stuff leads to a closer dependence on text encoder than usual with other models.

PR #1420, opened 2026-06-18
fix history search missing metadata for .webp images

jakstein:
So basically history panel's text search was missing images whose filenames don't contain the search term (e.g. searching by prompt content). The searchable string on each file is built from name, metadata, fullsrc, but .webp files always had null metadata because ImageSharp's WebP decoder doesn't populate ExifProfile.

The metadata IS embedded in webp files (as EXIF UserComment, written at ImageFile.cs:306), but GetMetadata() only checked img.Metadata.ExifProfile which is null for WebP in ImageSharp 3.1.12.

Changes:

OutputMetadataTracker.cs:123 - Add "webp" to ExtensionsWithMetadata
~~OutputMetadataTracker.cs:338 - Invalidate cached entries with null Metadata for extractable formats, so stale null-metadata entries (cached before webp was in the set) get force-re-read~~
~~ImageFile.cs:224 - Fallback to ReadExifUserCommentFromRaw() when ImageSharp's ExifProfile returns null~~
~~ImageFile.cs:238-370 - New ReadExifUserCommentFromRaw() method that parses the EXIF UserComment tag (0x9286) directly from raw binary bytes, matching what the client-side ExifReader JS library does successfully~~

### DIFF
--- a/src/Media/ImageFile.cs
+++ b/src/Media/ImageFile.cs
@@ -9,7 +9,6 @@ using SixLabors.ImageSharp.Formats.Png;
 using SixLabors.ImageSharp.Formats.Webp;
 using SixLabors.ImageSharp.Formats.Jpeg;
 using SixLabors.ImageSharp.PixelFormats;
-
 using Image = SwarmUI.Utils.Image;
 using ISImage = SixLabors.ImageSharp.Image;
 using ISImage32 = SixLabors.ImageSharp.Image<SixLabors.ImageSharp.PixelFormats.Rgba32>;
--- a/src/Utils/OutputMetadataTracker.cs
+++ b/src/Utils/OutputMetadataTracker.cs
@@ -120,7 +120,7 @@ public static class OutputMetadataTracker
     }
 
     /// <summary>File format extensions that even can have metadata on them.</summary>
-    public static HashSet<string> ExtensionsWithMetadata = ["png", "jpg"];
+    public static HashSet<string> ExtensionsWithMetadata = ["png", "jpg", "webp"];
 
     /// <summary>File format extensions that require ffmpeg to process image data.</summary>
     public static HashSet<string> ExtensionsForFfmpegables = ["webm", "mp4", "mov"];


src/Utils/OutputMetadataTracker.cs
```diff
@@ -336,24 +336,31 @@ public static OutputMetadataEntry GetMetadataFor(string file, string root, bool
             }
             if (existingEntry is not null)
             {
-                float chance = Program.ServerSettings.Performance.ImageDataValidationChance;
-                if (chance == 0 || Random.Shared.NextDouble() > chance)
+                if (existingEntry.Metadata is null && ExtensionsWithMetadata.Contains(ext))
```
mcmonkey4eva: all changes down here: no. Leave it as it was.

src/Media/ImageFile.cs
```diff
@@ -230,6 +235,141 @@ public string GetMetadata()
         return null;
     }
 
+    /// <summary>Fallback: reads EXIF UserComment directly from raw binary data when ImageSharp's ExifProfile is unavailable.</summary>
+    public string ReadExifUserCommentFromRaw()
```
mcmonkey4eva: This is deeply hackyweird. This should not be needed at all

src/Utils/OutputMetadataTracker.cs
jakstein: reverted

src/Media/ImageFile.cs
jakstein: turns out that this was indeed not needed, the bare minimum that works is adding the webp to ExtensionsWithMetadata, and (manual) removal of entire cache file. Rolling this back

mcmonkey4eva:
Lot more complicated than just adding to the list of allowed loaders, but I finally got around to bothering to try adding it and solving the side effects. Not sure if I've cured *every* side effect of the hell that is loading webps, but at least the surface level ones are good now.

PR #1424, opened 2026-06-22
Adds Krea2 support

jtreminio:
WIP

1744001-High definition photo of an orange cat


jtreminio:
CleanShot 2026-06-22 at 17 24 50@2x

PR #1438, opened 2026-06-29
Fixes *WS node previews

jtreminio:
The C# changes are necessary because comfy will only preview a video if it's saved to the tmp directory first... and I found that comfy lists that preview file in its history and swarm reads and pulls it in so the video ends up added twice.

So we teach swarm to ignore the history entry from any node that already sent its data over the websocket, and only care about the WS data when both are present.

CleanShot 2026-06-29 at 08 50 32@2x

Includes video support:

CleanShot 2026-06-29 at 08 21 42@2x

There is a small bug where if you gen an image first, then switch to a video model within swarmui and click Import From Generate Tab on the same tab, the previous image remains:

CleanShot 2026-06-29 at 08 20 30@2x


mcmonkey4eva:
This is the wrong approach, the file should not be touching the temp directory at least for images. Comfy's seminative `SaveImageWebsocket` uses websocket saves and previews correctly iirc

PR #1443, opened 2026-07-08
Adds conditioning multiply

jtreminio:


src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
```diff
@@ -2524,6 +2527,15 @@ public JArray CreateConditioningDirect(string prompt, JArray clip, T2IModel mode
                 ["clip"] = clip,
                 ["text"] = prompt
             }, id);
+            isPlainTextEncode = true;
+        }
+        if (isPlainTextEncode && UserInput.TryGet(T2IParamTypes.ConditioningMultiplier, out double condMultiplier) && condMultiplier != 1)
```
mcmonkey4eva: Silent exclusion for incompatibility image-input conds is awkward.
Also I feel like positive/negative prompt difference would be relevant here?
(Or I guess for Krea it doesn't really matter cause Turbo. but like. idk.)

Issue #1460, opened 2026-07-27
Krea 2 Edit Lora support


RaccoonOnGo:
### Feature Idea

Krea 2 edit lora's now work natively with ComfyUI, SwarmUI support would be great

https://github.com/Comfy-Org/ComfyUI/pull/14843

PR #1472, opened 2026-08-03
Add Krea2Edit Extension (#1460)

allenbenz:
Made an [extension](https://github.com/allenbenz/SwarmUI-Krea2Edit) that loops in the Krea2EditModelPatch to make things like https://huggingface.co/conradlocke/krea2-identity-edit work right and replicates what Krea2EditGroundedEncode does with swarm nodes so I think it might play nicer with other extensions/be composable.

I don't have good testing scenarios to verify how composable it is with other extensions so I don't really know.

It supports both one and two input images and exposes all the options that are available on Krea2EditModelPatch (ref boost/ref boost a/legacy crop) and exposes would have been available if Krea2EditGroundedEncode was being used (grounding_px).

It pulls in https://github.com/lbouaraba/comfyui-krea2edit and https://github.com/cubiq/ComfyUI_essentials, the latter is a old and in maintenance mode but it has the image resize node that can replicate how Krea2EditGroundedEncode resizes images.

This copies liberally from KreaGoneWild.

### DIFF
--- a/launchtools/extension_list.fds
+++ b/launchtools/extension_list.fds
@@ -367,3 +367,12 @@ KreaGoneWild:
     tags:
     - parameters
     - nodes
+
+Krea2Edit:
+    author: Allen Benz
+    description: Adds support for models that require the ComfyUI-Krea2Edit nodes.
+    url: https://github.com/allenbenz/SwarmUI-Krea2Edit
+    license: MIT
+    tags:
+    - parameters
+    - nodes


mcmonkey4eva:
I believe Comfy added upstream core support for Krea ref images, so using the old external custom node packs is probably a bad idea.

Instead we should update Swarm to support prompt images for Krea (and some way of routing between the different krea edit models, not sure how relevant that is)

allenbenz:
Okay, thanks for looking at it.

Leaving a note for myself or anyone else looking into this.

https://github.com/Comfy-Org/ComfyUI/pull/14843 is what got merged upstream, for osiris it looks fine but for the identity edit v1.2 route it's missing ref_boost and identity edit needs that to be useful and the RoPE offset is 0,0 instead of centered which doesn't break the model but shifts things over to the left when the aspect ratio of the input and the output differ, maybe doesn't play well with two inputs, didn't look into that.

PR #1481, opened 2026-08-06
Add "Custom" option to Video Resolution param

grossi:
Currently none of the Video Resolution modes give direct control over the video stage's output resolution: "Model Preferred" is fixed to the model's standard res (e.g. always 960x960 for MiniMax H3), "Image Aspect, Model Res" is locked to the model's pixel budget, and "Image" follows the input image resolution and is also multiplied by Refiner Upscale when that's set.

This adds a "Custom" option that uses the standard Width/Height parameters directly for the video, rounded to the video model's `ResolutionPrecision` (e.g. x32 for MiniMax H3). Useful for trading resolution against frame count on VRAM-limited setups.

Default behavior is unchanged. Tested with MiniMax H3 FL2VA image-to-video on an RTX 4080.


mcmonkey4eva:
... Wut? Use `Image`

brendanhoar:
Let me suggest a use case.

When performing T2I2V, and already using the global aspect/side (or custom resolution) to control the Image output size, say, for an Image model that doesn't handle low resolutions well, but the video model does handle low resolutions well and you would like to constrain the size / VRAM / time of the video generation, this is exacty where being able to insert a custom resolution (and/or or scale factor) target at the video stage would actually be quite useful.

mcmonkey4eva:
This PR doesn't do that

grossi:
This PR uses the value from the Image.

My bad, I thought that the "Image" option used the input image's resolution. I agree this PR should be closed.

PR #1484, opened 2026-08-06
Adds a per-generation attention mode override dropdown (disabled/sdpa/pytorch/sage/flash), letting you opt specific models out of global SageAttention without restarting ComfyUI.

barelymining:
Adds the SageAttentionOptout extension which wraps the comfy-sageattention-optout-customnodes node by EnviralDesign to allow per-generation attention mode overrides without restarting ComfyUI.

### DIFF
--- a/launchtools/extension_list.fds
+++ b/launchtools/extension_list.fds
@@ -367,3 +367,14 @@ KreaGoneWild:
     tags:
     - parameters
     - nodes
+
+SageAttentionOptout:
+    # Wraps https://github.com/EnviralDesign/comfy-sageattention-optout-customnodes by EnviralDesign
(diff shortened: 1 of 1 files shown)


barelymining:
Some models don't play nice with SageAttention so this is really useful!
image

FurkanGozukara:
this gave me nice idea i should make this

barelymining:
not sure what went wrong during the request, anything I should try? @mcmonkey4eva

mcmonkey4eva:
There's now a core feature to do this, and it provides Comfy's new native "comfy kitchen attention" which is essentially sage attention but slightly better and much easier to get working

PR #1496, opened 2026-08-12
Fix edit model metadata linked preset focus steal

jtreminio:
Bug fix where clicking the Linked Preset dropdown (`(None)` in screenshot) auto-loses focus to the model name field and makes it impossible to type into the search bar.

download


mcmonkey4eva:
fixed more correctly in separate commit

PR #1505, opened 2026-08-16
Add ModelPaths extension to the list

Janush1981:
Adds [SwarmUI-ModelPaths](https://github.com/Janush1981/SwarmUI-ModelPaths) to the
known extensions list.

**What it does:** Swarm maps eight configurable model path types plus a short hardcoded
list in `FoldersToForwardInComfyPath`. Everything else — `LLM`, `insightface`,
`ultralytics`, `sam`, `mmaudio`, `florence2`, `rembg` and so on — never reaches the
generated comfy paths file, so nodes fall back to `ComfyUI/models/` on the install drive
and re-download models that already exist on the model root. On my machine that had
silently duplicated 62 GB.

The extension discovers folders instead of taking a hand-written list: whatever exists
under `Paths.ModelRoot` is forwarded. No new setting, no filesystem scan — it lists the
immediate subdirectories of the root Swarm already required, one level deep.

It also merges duplicate folder-type keys via the `ModifyComfyYaml` hook. Nothing orders
extension init, so two extensions can claim the same key without seeing each other — I hit
this with `luts`, wanted by both PostRenderTorched and folder discovery. YAML resolves a
duplicate key last-one-wins, silently dropping the other's paths.

MIT licensed, `ci-test: true`, README in English and Ukrainian. Known limitations are
documented rather than glossed over: nodes that hardcode `models_dir` (`ace-step`) or
overwrite `folder_names_and_paths` outright (ReActor) bypass path registration entirely,
and no amount of registering fixes that.

### DIFF
--- a/launchtools/extension_list.fds
+++ b/launchtools/extension_list.fds
@@ -376,3 +376,12 @@ H3Attn:
     - parameters
     - nodes
     - beta
+
+ModelPaths:
+    author: Janush1981
+    description: Forwards every model folder under your model root to ComfyUI, not just the fixed set Swarm maps - LLM, insightface, ultralytics, sam, mmaudio, florence2 and the rest - so custom nodes stop re-downloading models you already have.
+    url: https://github.com/Janush1981/SwarmUI-ModelPaths
+    license: MIT
+    ci-test: true
+    tags:
+    - tools


mcmonkey4eva:
(1) this is pure claudeslop
(2) this is not appropriately relevant for an extension to be doing vs a core change
(3) this doesn't look even look functionally correct at a quick glance through

Janush1981:
Thanks for the slop callout 🙂 and for actually reading it. You were right about (3).

Straight up: I'm not a strong C# dev. I had 62 GB of duplicated models on my install drive, went digging for why, and ended up writing something. That's the honest origin story.
The bug — the merge treated the yaml as one flat namespace. Core writes a section per model root, each with its own base_path, and the same folder keys repeat across them. So with two roots the second one lost every key the first also had: `swarmui2` came out with no `checkpoints`, no `loras`. Keyed it on (section, key) now. Doesn't change
your point (2).
What's underneath is what I actually care about. `FoldersToForwardInComfyPath` is a fixed list, so `LLM`, `insightface`, `ultralytics`, `sam*`, `mmaudio`, `florence2`, `rembg` and friends never reach the generated paths file. Nodes don't find them and pull their own copy — I watched ReActor do exactly that, 1.3 GB, seconds after I'd cleaned the folder out. 62 GB of duplicates, all of it already sitting on the model root.
You said core, not extension. You know this codebase, and I don't, so I'll take your word for it. What would you actually take, if anything? Auto-forward every subfolder of `ModelRoot`, a curated list, something else? Or if it's already on your pile, say
so and I'll drop it and stop making noise.
One thing I tripped over that might be worth knowing anyway: nothing dedupes `FoldersToForwardInComfyPath`, so two extensions can claim the same folder type and the YAML gets a duplicate key. YAML takes the last one; the other's paths vanish. Hit it with `luts` — PostRenderTorched and mine both wanted it. Maybe old news to you. 
I'll just fix my slop for my one usage ;)

PR #1532, opened 2026-09-05
Adds parallel downloader

jtreminio:
Parallel downloader for HF files, with a fallback to the linear downloader.

~Shares as much existing functionality between the two as I could squeeze in.~ _edit_: reverted.

Very scientific comparisons:

* ~55mb/s without fix
* ~320mb/s with fix


mcmonkey4eva:
Ended up redoing the concept myself in a separate commit

PR #1539, opened 2026-09-12
Adds YuE2 support

jtreminio:
Need guidance on how you want (if you want) to expose mode and max_abc_tokens

* https://github.com/Comfy-Org/ComfyUI/pull/16250
* https://huggingface.co/Comfy-Org/YuE2

Model needs to go in Stable-Diffusion folder

### DIFF
--- a/src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
+++ b/src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
@@ -2588,7 +2588,7 @@ public partial class WorkflowGenerator
                 ["abc"] = NodePath(abc, 0),
                 ["seed"] = UserInput.Get(T2IParamTypes.Seed, 0) + 20,
                 ["mode"] = "full",
-                ["max_duration"] = Math.Clamp(UserInput.Get(T2IParamTypes.Text2AudioDuration, 120), 0.04, 360),
+                ["max_duration"] = Math.Clamp(UserInput.Get(T2IParamTypes.Text2AudioDuration, 120), 0.04, 900),
                 ["temperature"] = 1,
                 ["top_p"] = 0.95,
                 ["top_k"] = 100,
--- a/src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
+++ b/src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
@@ -1014,6 +1014,11 @@ public partial class WorkflowGenerator
             defsampler ??= "res_multistep";
             defscheduler ??= "simple";
         }
+        else if (IsYue2())
+        {
+            defsampler ??= "dpm_2";
+            defscheduler ??= "sgm_uniform";
+        }
         else if (IsAnima())
         {
             defsampler ??= "er_sde";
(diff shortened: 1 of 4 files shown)


src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
```diff
@@ -1014,6 +1014,11 @@ public string CreateKSampler(JArray model, JArray pos, JArray neg, JArray latent
             defsampler ??= "res_multistep";
             defscheduler ??= "simple";
         }
+        else if (IsYue2())
+        {
+            defsampler ??= "dpm_2";
+            defscheduler ??= "sgm_uniform";
```
jtreminio, line 1020: honestly euler/normal is perfectly cromulent

src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
```diff
@@ -2560,6 +2565,36 @@ public JArray CreateConditioningDirect(string prompt, JArray clip, T2IModel mode
...
+            {
+                return FinalPrompt;
+            }
+            string abc = CreateNode("YuE2GenerateABC", new JObject()
+            {
+                ["clip"] = clip,
+                ["style"] = UserInput.Get(T2IParamTypes.Text2AudioStyle, ""),
+                ["lyrics"] = prompt,
+                ["seed"] = UserInput.Get(T2IParamTypes.Seed, 0) + 10,
+                ["mode"] = "full",
```
jtreminio, line 2580: full or melody

src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
```diff
@@ -2560,6 +2565,36 @@ public JArray CreateConditioningDirect(string prompt, JArray clip, T2IModel mode
                 ["min_p"] = 0
             }, id);
         }
+        else if (IsYue2())
+        {
+            if (!isPositive)
+            {
+                return FinalPrompt;
+            }
```
jtreminio, line 2573: I noticed if you create duplicate YuE2GenerateABC/YuE2GenerateMusic and pipe to ConditioningZeroOut for the negative prompt, comfy helpfully duplicates the work it does, even when all params are identical.

Could have sworn identical nodes would be no-ops as their output would have been cached, but that wasn't the case here.

mcmonkey4eva: That's swarm side actually - Swarm's addnode will check for duplicates and just not add duplicate nodes - if you're getting actual multiple nodes in the final workflow, then they're not identical

src/BuiltinExtensions/ComfyUIBackend/WorkflowGenerator.cs
```diff
@@ -2560,6 +2565,36 @@ public JArray CreateConditioningDirect(string prompt, JArray clip, T2IModel mode
...
+                ["style"] = UserInput.Get(T2IParamTypes.Text2AudioStyle, ""),
+                ["lyrics"] = prompt,
+                ["abc"] = NodePath(abc, 0),
+                ["seed"] = UserInput.Get(T2IParamTypes.Seed, 0) + 20,
+                ["mode"] = "full",
+                ["max_duration"] = Math.Clamp(UserInput.Get(T2IParamTypes.Text2AudioDuration, 120), 0.04, 900),
+                ["temperature"] = 1,
+                ["top_p"] = 0.95,
+                ["top_k"] = 100,
+                ["repetition_penalty"] = 1.2
```
mcmonkey4eva, line 2595: Where did these values come from? They don't match comfy's reference or the node defaults

